Issue #30, opened 2021-04-26
Navigation to scroll independently of content


alexroan:
The left menu is anchored to the main content, in that to scroll one; you must scroll both.

This means that the length of the page is dictated by the longest section (either menu or content). When you land on a page, you land at the top, which is desired for the main content. However, it is not for the menu, where the active menu item is often further down the page.

Some ideas:

- [x] Have the height of the left menu navigation container be determined by the length of the main content.
- [x] Make the menu itself scrollable, independent of the main content scroll.
- [ ] Auto-scroll the menu on page-load to vertically centre the active item.

PR #37, opened 2021-04-27
Make navigation scrollable

dmitrydao:


StephenFluin (dismissed):
This seems to break the ability to resize the page. Now the menu is collapsing prematurely. See screenshot.

localhost_4200_ (1)

_src/docs.css
```diff
@@ -121,9 +121,14 @@ a:hover,
 }
 navigation {
```
alexroan, line 122: Referencing #30, could we auto-scroll the menu to the active item?

alexroan (changes requested):

_src/docs.css
dmitrydao: Can you clarify?

StephenFluin: @alexroan Can we land this as a separate feature / PR?

alexroan (approved):

PR #79, opened 2021-05-10
renamed function names in Any-API code example

pappas999:
function names were for the old request ethereum price API call.

Take note, the gist e5f752fa134b49ef481da74ec1a453a6 needs updating too (which I don't have access to)


StephenFluin:
Can we use a new folder in this repo instead of a gist? We want to move them all towards being integrated.

dmitrydao:
> Can we use a new folder in this repo instead of a gist? We want to move them all towards being integrated.

I like the idea, it would be easier to manage code samples. The only problem is that Remix relies on gists to open code samples.

dmitrydao:
> function names were for the old request ethereum price API call.
> 
> Take note, the gist e5f752fa134b49ef481da74ec1a453a6 needs updating too (which I don't have access to)

Ouch. Did that code sample a couple of months ago and didn't even realize the name of the function didn't match with what it represents. That was also my gist, already updated.

alexroan:
@dmitrydao & @pappas999 , what's the status on this?

dmitrydao:
> @dmitrydao & @pappas999 , what's the status on this?

I still haven't figured out how to embed code samples from local files to open them in the remix. Not sure that's currently possible.

pappas999:
@alexroan what Stephen proposed is absolutely something that should be done, but currently isn't supported by Remix. So for now I think PR is ready to move forward in it's current form

alexroan (approved):

PR #83, opened 2021-05-13
updated for chainlink spring 2021 winners

PatrickAlphaC:


StephenFluin (approved):
Long term we want to move away from `a` tags towards markdown. But this matches the existing style of the page so it's okay for now.

PR #84, opened 2021-05-13
Add additional info section

dmitrydao:
## Overview

Add a common section `Additional Information` with missing pages: 
- FAQ
- Glossary
- Workshop Links
- Example Projects


StephenFluin:
Some of the information on the FAQ is out of date. If we're re-adding links to it, I think we should probably update it at the same time.

dmitrydao:
> Some of the information on the FAQ is out of date. If we're re-adding links to it, I think we should probably update it at the same time.

What is the info that seems outdated? I can only see boxes (Truffle, Hardhat, Brownie) that have a different naming here.

StephenFluin:
@dmitrydao Can you makes sure we are moving forward to a next step on this one?

alexroan:
@dmitrydao bump

PR #131, opened 2021-06-05
Update docs for finding node address

cds95:
I noticed that the docs for finding the node address in the node UI is out of date.  Later versions of the node seem to no longer display the node address in the configurations page but instead displays them in the Account Address section at the bottom of the 'Keys' page.  I'm not sure what's the earliest node version that has this change but I think that information would be useful to add if it is known.

New Doc Layout.
Screen Shot 2021-06-05 at 2 20 22 PM


StephenFluin:
This change is great! Do you think you could compress the image (using something like [trimage](https://trimage.org/)) to get the filesize down? Also, if you can host the file in `_src/images/node-operators/`?

cds95:
hi @StephenFluin sorry I missed your message.  I've compressed the image and moved it to the folder you asked.  Let me know if you need anything else in this MR.

StephenFluin:
This has been merged with a fixup.

PR #140, opened 2021-06-10
Fix images and hidden properties

dmitrydao:


PatrickAlphaC:
Do we all need to be requested reviewers on these?

dmitrydao:
> Do we all need to be requested reviewers on these?

I think it automatically picks you up as you're the code owner.
Will fix that building error soon, didn't notice it failed.

PatrickAlphaC:
> > Do we all need to be requested reviewers on these?
> 
> I think it automatically picks you up as you're the code owner.
> Will fix that building error soon, didn't notice it failed.

Ok, I'll have to filter out my emails. Thanks

StephenFluin (changes requested):
looking really great, can you replace the 7 readme image blocks with pure markdown?

StephenFluin (approved):

PR #148, opened 2021-06-17
Update data providers list

dmitrydao:


StephenFluin:
Have we considered removing this list and pointing folks to market.link?

dmitrydao:
> Have we considered removing this list and pointing folks to market.link?

That's exactly what we did but there are some data providers that are not on market.link so they should be covered as well. If they're not used by anyone, I'm up for removing the list entirely.

StephenFluin (approved):

StephenFluin (changes requested):
Can we remove the pages we aren't linking to anymore, and add forwards so that users end up on the right place and we don't have two sources of truth?

StephenFluin (changes requested):
The removed files look good but we still need to add 301 redirects. Can you do that in the firebase.json file?

StephenFluin (approved):

PR #150, opened 2021-06-17
Changed Core Adapters primary URL

ZakAyesh:
Put old URL in firebase.json so backlinks still work.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -4,12 +4,12 @@
     "ignore": ["firebase.json", "**/.*", "**/node_modules/**"],
     "redirects": [
       {
-        "source": "/docs",
+        "source": "/docs/",
         "destination": "/",
         "type": 301
       },
       {
-        "source": "/docs/getting-started",
+        "source": "/docs/getting-started/",
         "destination": "/",
         "type": 301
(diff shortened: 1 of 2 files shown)


firebase.json
```diff
@@ -22,6 +22,11 @@
         "source": "/docs/kovan-keeper-network-beta/",
         "destination": "/docs/chainlink-keepers/introduction/",
         "type": 301
+      },
+      {
+        "source": "/docs/adapters",
```
StephenFluin, line 27: Can you make sure these end with `/`?

StephenFluin (changes requested):

StephenFluin:
It looks like there are 3 commits right now. Can you squash them and force push the PR so there's only one that's based off of master?

firebase.json
ZakAyesh: Completed and squashed changes into one commit.

ZakAyesh:
> It looks like there are 3 commits right now. Can you squash them and force push the PR so there's only one that's based off of master?

Should be all good now

StephenFluin (approved):

PR #169, opened 2021-06-23
add scheduled link checking to CI

StephenFluin:
Starting with a scheduled job.

We should be checking this as a standard CI check for every PR, but it's not clear how to spin up a local instance in a clean way.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -1,0 +1,20 @@
+name: Check Docs and Site Links
+
+on:
+  schedule:
+    - cron: '30 8 * * *'
+
+jobs:
+  check-links:
+    name: Regularly check links
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


dwightjl (approved):
Looks good to me. Merge when appropriate.

.github/workflows/check-links.yml
```diff
@@ -0,0 +1,19 @@
...
+  schedule:
+    - cron: '30 8 * * *'
+
+jobs:
+  check-links:
+    name: Regularly check links
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check docs links with linkcheck
+        uses: filiph/linkcheck@v2.0.15+1
```
dwightjl, line 13: 2.0.19 is available if we want to move up to that.
https://github.com/filiph/linkcheck/releases/tag/2.0.19

PR #174, opened 2021-06-24
update kovan faucet to protofire

PatrickAlphaC:


StephenFluin:
The new faucet offers LINK and ETH. Where should we point people to for LINK? We currently have this and the old LINK faucet for Kovan?

StephenFluin (approved):

StephenFluin:
Approving with the two faucets.

PR #176, opened 2021-06-24
Feat/update kovan faucet

PatrickAlphaC:


StephenFluin (approved):

StephenFluin:
I wish these instructions said to paste your address, but we can improve that later.

StephenFluin:
Actually, I just noticed that this PR now has 3 commits. Can you squash them and rebase on main into one so we don't have the merge commit and so we have a clean history?

PatrickAlphaC:
@StephenFluin all set. And added pasting address.

PR #178, opened 2021-06-28
Feed Registry documentation

yosriady:
> [About the feed registry](https://docs.google.com/presentation/d/1p-ZNM2cwNEzgcexOsqRMQh4vr3al-Xl1mtemFJJ-eN4/edit) (pitch deck)

This PR adds documentation for the upcoming Feed Registry https://github.com/smartcontractkit/feed-registry, initially available for preview users.

Screenshot 2021-06-28 at 2 56 53 PM


yosriady:
TODOs

- [x] Update alpha definition
- [ ] Wait for the new contracts to be published to `@chainlink/contracts` before merging
- [ ] Update code examples to support v0.6/v0.7/v0.8 tabs
- [ ] Prepare a runnable Remix example

yosriady:
@StephenFluin This is ready for another review!

ZakAyesh:
@yosriady 
Took a look over and here are a couple of items to be addressed:

- Make sure to make this a single commit before it is merged so we don't bloat the main branch commit-tree
- Can’t use the registry on Kovan, we shouldn’t include the Remix gist example code until it we have the registry on at least Kovan
- Instead of their own hardhat project can we integrate it with Chainlink-HardHat-StarterKit?
- Be clear that the LINK token addresses given are for main-net
- The page includes a link to the API Reference and also includes that page explicitly. All’s we need is the link to the page.

PatrickAlphaC:
@ZakAyesh, actually, I like the standalone project **and** the integration with the starter kit. 
Unless we think it's too much to maintain.

yosriady:
@ZakAyesh @PatrickAlphaC Do we have 'Squash merge' enabled on this repo? Can we set that as a default?

yosriady:
@ZakAyesh 

> The page includes a link to the API Reference and also includes that page explicitly. All we need is the link to the page.

If you look closely, the Feed Registry API reference is different. It uses additional parameters such as `asset` and `denomination`. There are also additional functions available on `FeedRegistryInterface` that could be added here, but omitting it for now for conciseness.

yosriady:
TODOs
- [ ] Deploy Feed Registry + generate tooling + execute addaccess multisig proposal on Kovan OR remove Remix link for now
- [ ] Add to hardhat-starter-kit

yosriady:
@ZakAyesh @PatrickAlphaC Setting up the Kovan registry will take time due to the need to create and execute multisig proposals to grant access from the aggregators to the registry. 

Can we just remove the Remix link the `hardhat-starter-kit` additions since both wants to use Kovan, and publish what we have so far? After the Kovan registry is deployed and set up, I'll create another PR to add the new Remix link for Kovan.

ZakAyesh:
@yosriady 
>Should this be a separate API Reference page? I figured having it on the same page makes the most sense.

Thanks for clarifying! I think keeping the API reference is fine while the feature is in alpha/beta. After it is a full release we can think about separating it out.

> @ZakAyesh @PatrickAlphaC Can we omit the mainnet Remix link for now & defer the `hardhat-starter-kit` additions so we can first publish what we have so far on Mainnet?

Sounds good. 

I made one small typo fix, PR looks good to go.

PatrickAlphaC (approved):

PR #187, opened 2021-06-30
Minor rewrites, fixes, and convert callouts to markdown.

dwightjl:
View these changes on a temporary staging server: http://home.taters.org/docs/running-a-chainlink-node/

- Remove extra `\` characters from `echo` command examples. These were left over from converting tabbed code blocks to markdown.
- Fix a broken `docker inspect --format` example. The format contents disappeared because they were in `{{ }}` brackets, which Liquid interpets as Liquid tags even within markdown codeblocks. See the currently published example: https://docs.chain.link/docs/running-a-chainlink-node/#ethereum-client-on-the-same-machine
- Rewrite the "Set your Ethereum Client URL" section to be more simple.
- Rewrite the requirements section to include network requirements that were previously documented in a callout box.
- Convert callout boxes to markdown format.


StephenFluin (changes requested):
I'm not sure if this is intentional or not, but we lost the # Using Docker header and recommendation. We definitely want to keep that.

dwightjl:
I messed up the squash somehow. Retrying with a fresh request.

PR #191, opened 2021-07-01
add link to vrf best practices link

### LINKED ISSUE #183: add link to new VRF best practices
When we removed the VRF randomness expansion tip and moved it to the new section, some folks have been trouble finding it.

Let's add a link where the old description was to the new page & section.

https://github.com/smartcontractkit/documentation/pull/138/files#diff-7f4445627191c23a2b102a6d4ae6ec0991d4a16ae0cb91e2018099769b038cddL100-L114

dmitrydao:
fixes #183


StephenFluin (changes requested):
This isn't quite right. The problem is that we used to list "VRF randomness expansion" here, so people are looking for that. The text should say something like "Getting More Randomness" "If you are looking for how to turn a single result into multiple random numbers, check out our guide on Randomness Expansion" and link DIRECTLY to the VRF randomness expansion section.

StephenFluin (approved):

PR #194, opened 2021-07-02
remove duplicate values in redirect settings

dmitrydao:


StephenFluin:
These don't look like duplicates to me? These seem to handle both the slash and missing slash?

dmitrydao:
> These don't look like duplicates to me? These seem to handle both the slash and missing slash?

Oh. Didn't notice that. Closing

PR #205, opened 2021-07-07
autoscroll navigation and fix subscribe feature

StephenFluin:


StephenFluin:
@alexroan Can you take a look? <3

_src/scroll.js
```diff
@@ -0,0 +1,8 @@
+document.addEventListener('DOMContentLoaded', () => {
+    const element = document.querySelector('navigation .active');
```
alexroan, line 2: Instead of auto-scrolling the active menu item to the top of the window, could we scroll it to the centre?

alexroan (changes requested):

_src/scroll.js
StephenFluin: This is what should happen on most browsers: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded#browser_compatibility

What browser are you seeing this on?

alexroan: I'm using Chrome, and the active menu item is being scrolled to the top of the window.

StephenFluin:
@alexroan take another look, flipped the conditional check to fix the bug, and added resizing of the sidebar so it's always the right size.

alexroan (approved):

PR #227, opened 2021-07-16
https://github.com/smartcontractkit/chainlink/blob/master/evm-contrac…

toeydaow:
…ts/src/v0.6/interfaces/AggregatorV3Interface.sol

https://docs.chain.link/docs/price-feeds-api-reference/


StephenFluin:
It looks like this PR just adds a newline at the end of the file. We generally fix things like this when other changes are going into the file, so I'll close the PR for now.

Thanks for the contribution!

PR #251, opened 2021-07-22
remove escape character in shell samples

### LINKED ISSUE #244: Shell commands are not copy paste compatible - remove escape character
For example: 

```
https://docs.chain.link/docs/running-a-chainlink-node/#create-an-environment-file
```

When you copy paste the text from:
```
echo \"ROOT=/chainlink
LOG_LEVEL=debug
ETH_CHAIN_ID=1
CHAINLINK_TLS_PORT=0
SECURE_COOKIES=false
GAS_UPDATER_ENABLED=true
ALLOW_ORIGINS=*\" > ~/.chainlink/.env
```
and paste that into your shell, it will execute each line independently.

It should be:
```
echo "ROOT=/chainlink
LOG_LEVEL=debug
ETH_CHAIN_ID=1
CHAINLINK_TLS_PORT=0
SECURE_COOKIES=false
GAS_UPDATER_ENABLED=true
ALLOW_ORIGINS=*" > ~/.chainlink/.env
```

We need to get rid of the escape character

dmitrydao:
fixes #244

### DIFF
--- a/docs/Node Operators/enabling-https-connections.md
+++ b/docs/Node Operators/enabling-https-connections.md
@@ -62,7 +62,7 @@ Next, add the `TLS_CERT_PATH` and `TLS_KEY_PATH` environment variables to your `
 {
   "codes": [
     {
-      "code": "echo \"TLS_CERT_PATH=/chainlink/tls/server.crt\nTLS_KEY_PATH=/chainlink/tls/server.key\" >> .env",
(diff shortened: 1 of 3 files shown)


StephenFluin (changes requested):
I don't think this is right. For any [block:] style JSON, the escape characters need to stay. It's only when we move to backtick notation (which we should do for all of these examples) that you need to remove the escape characters.

For example, on enabling-https, the echo command looks right in the docs today. Feels like a lucky bug that removing the escape chars still works in those JSON examples.

dmitrydao:
Closing PR. Replaced with #258

PR #252, opened 2021-07-22
add heco to the list of supported chains

### LINKED ISSUE #238: HECO is not available in https://docs.chain.link/docs/link-token-contracts/
I am not sure if HECO has no testnet, but in case it does maybe good to add it. The mainnet is also not listed https://docs.chain.link/docs/link-token-contracts/.

dmitrydao:
fixes #238


StephenFluin (changes requested):
Why is the address empty / `?`?

dmitrydao:
> Why is the address empty / `?`?

Because I don't know the address and I'm waiting for access to the repo that will give me the answers for this.
That's why it's WIP

StephenFluin:
I missed that this was a draft. Perfect! Sorry!

StephenFluin (approved):

PR #264, opened 2021-07-25
Update example in the "Architecture decentralized data model" section of the docs

Sylvarant92:
Update the number of oracles in the docs from 21 oracles to 31 oracles for the ETH/USD feed example. Current ETH/USD price feed has 31 oracles actively serving data requests.
Source: https://data.chain.link/ethereum/mainnet/crypto-usd/eth-usd


dmitrydao:
I think it doesn't make sense to update the number of oracles because the number will always fluctuate. You can add something like this instead:

The number of oracles is not fixed and will vary as more nodes join the network.

StephenFluin:
Closing this PR as we should make the text less specific rather than updating it. Feel free to send another PR if you are interested in making this change, and I'm sorry we couldn't accept this one.

PR #276, opened 2021-08-04
Chainlink Keepers - Update register-upkeep.md - adding section for Maintaining a minimum balance

AnushV:
Providing more context for how to think about minimum balance.

### DIFF
--- a/docs/chainlink-keepers/register-upkeep.md
+++ b/docs/chainlink-keepers/register-upkeep.md
@@ -64,6 +64,13 @@ The App will walk you through several steps.
 * Anyone (not just the Upkeep owner) can call the `addFunds()` function
 * To withdraw funds, first cancel the Upkeep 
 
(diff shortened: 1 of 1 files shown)


StephenFluin:
Fixing typos and then going to force push and approve.

> ## Maintaining a minimum balance
> To ensure the Chainlink Keepers are compensated for performance, there is an expected minimum balance on each Upkeep.
> If your funds drop below this amount, the Upkeep will not be performed.
> The minimu balanace is calculated using the current fast gas price, the Gas Limit you entered for your Upkeep, and the max gas multiplier (see `gasCeilingMultiplier` in [configuration of the registry](/docs/chainlink-keepers/overview/#configuration)).

> In order to ensure performance, it is recommended that you maintian a balance that is multiple (3-5x) the minimum balance to account for gas price fluctuations.

StephenFluin (approved):

PR #284, opened 2021-08-06
Specify call frequency to checkUpkeep function

cenobyte321:
According to the OpenZeppelin guide, the `checkUpkeep` function "will be called every block (about ~15 sec)". https://docs.openzeppelin.com/defender/guide-chainlink#implementing_the_upkeep_interface. I propose to specify the frequency in the Keepers docs.


StephenFluin:
Thank you so much for this contribution! We can't accept this at this time, as the text is explicitly written NOT to be every block. There are cases where this won't be true, such as on other non-ethereum chains.

cenobyte321:
@StephenFluin Thank you for reviewing this and for clarifying this point. Maybe there could be another way to convey this info in the docs? 🤔. I had this doubt the first time I looked into Keepers, and I've seen other people have had the same question in Discord.

PR #287, opened 2021-08-06
added information for gcp-weather node

PatrickAlphaC:


docs/Data Provider Nodes/google-weather.md
```diff
@@ -0,0 +1,361 @@
+---
+layout: nodes.liquid
+section: smartContract
+date: Last Modified
+title: "Google"
```
StephenFluin, line 5: Could we make the title more specific about what this page is?

StephenFluin (changes requested):
Should this be listed on the navigation or the list at https://deploy-preview-287--dreamy-villani-0e9e5c.netlify.app/docs/data-provider-nodes/ ?

bryanjowers:
Looking at preview, looks good. Do we still have a plan for indicating that this kind of data is only sourced from a single data provider / a single Oracle that is providing responses?

docs/Data Provider Nodes/google-weather.md
PatrickAlphaC: Wow, totally spaced on that. Thank you!

PatrickAlphaC:
@bryanjowers this doc is in the data node provider section of the docs. We used to have a section that clarified this, although it can be easily implied by the descriptions as well. I'll defer to @StephenFluin on why we took them out in the first place.

docs/Data Provider Nodes/google-weather.md
PatrickAlphaC: And yes, it should be in that section!

PatrickAlphaC:
@bryanjowers the section here: https://docs.chain.link/docs/data-provider-nodes/ explains the function of these types of oracles

StephenFluin (approved):

PR #229, opened 2021-07-16
Update docs to deprecate V1 and add basic V2 job

samsondav:

PR #292, opened 2021-08-11
Stephen/v1 jobspec deprecation

StephenFluin:
@spooktheducks 

This is a new PR with the 12 commits from https://github.com/smartcontractkit/documentation/pull/229 squashed and then all of my changes as a commit on top of it. I think this is ready to go. Take a look. If there aren't any objections or discussions, then I'll squash my commit into Sam's and we can get this merged ASAP.


StephenFluin:
@PatrickAlphaC can I can an approval from you? I believe @spooktheducks has already verbally approved this.

PatrickAlphaC:
These are beautiful.

I'd just add that we've removed the "initiator" concept in favor of just "job type".

Issue #310, opened 2021-08-18
Move Build, Deploy, and Sell Your Own Dynamic NFT under VRF


dwightjl:
Modify https://github.com/smartcontractkit/documentation/blob/main/docs/Introduction/tutorials/developer-tutorials.md

Move the link for "Build, Deploy, and Sell Your Own Dynamic NFT" to the VRF section.

dwightjl:
Look at changing the nav naming.

PR #312, opened 2021-08-18
Move "Build, Deploy, and Sell Your Own Dynamic NFT" to the VRF section.

dwightjl:
Issue https://github.com/smartcontractkit/documentation/issues/310
Move "Build, Deploy, and Sell Your Own Dynamic NFT" to the VRF section, which might be a more applicable section for this content.


_data/navigation.js
```diff
@@ -162,8 +162,8 @@ module.exports = {
                 url: '/docs/advanced-tutorial/',
               },
               {
-                title: 'Developer Tutorials',
-                url: '/docs/developer-tutorials/',
+                title: 'Tutorial Reference',
```
dwightjl, line 165: I'm not completely satisfied with "Tutorial Reference" as a title, but for now that's what I suggest.

StephenFluin: What about "Other Tutorials"?

StephenFluin: I'm also not certain it deserves top billing in the introduction section, vs the reference sections lower in the navigation.

dwightjl: The challenge here is that we have only one Tutorials section and right now its in our introduction. Long-term we might distinguish between "Getting Started" and "Tutorials", which would resolve the question of where this content should live in the navigation.

"Other Tutorials" is fine as a title for now. It flows acceptably out of the beginner tutorials, I think.

dmitrydao (approved):
LGTM

PR #253, opened 2021-07-22
move code samples to a single location

dmitrydao:

PR #316, opened 2021-08-19
rewrite deploy URLs for preview domains

StephenFluin:
Merge this after #253 is merged.


_src/deploy-to-github.js
```diff
@@ -0,0 +1,14 @@
+/** Fixes deployment URLs when on a preview URL */
+document.addEventListener('DOMContentLoaded', () => {
+  const host = window.location.hostname;
+  if (host !== 'docs.chain.link') {
+    // Rewrite Remix URLs with current hostname
+    // eg https://remix.ethereum.org/#version=soljson-v0.6.7+commit.b8d736ae.js&optimize=false&evmVersion=null&url=https://docs.chain.link/samples/VRF/VRFD20.sol
+
+    for (let item of document.links) {
+      if (item.href.startsWith('https://remix.ethereum.org')) {
+        item.setAttribute('href', item.href.replace('docs.chain.link', host));
```
dwightjl, line 10: The way we currently have `yarn serve` configured run Browsersync, it uses http by default instead of https. Forcing `https://` into the browser URL just returns `ERR_SSL_PROTOCOL_ERROR`.

Can you modify this to detect and apply `http://` or `https://` depending on what the host is using? Either that or we need to set up Browsersync to just redirect.

dwightjl (changes requested):

_src/deploy-to-github.js
dwightjl: I guess we're not specifically worried about local testing, which likely doesn't have ports 443 or 80 open externally anyway. Just good to remember that some Remix links won't work when using `yarn serve`.

dwightjl (approved):

StephenFluin:
@dwightjl Perhaps we want to detect localhost and NOT replace the docs.chain.link URL? This would mean that any already-existing code samples would keep working, but new/changed ones would not work or be updated with `yarn serve`. Not sure how to fix this because the code only exists on your machine, and is not accessible to remix.

dwightjl:
> rhaps we want to detect localhost and NOT replace the docs.chain.link URL? This would mean that any already-existing code samples would keep working, but new/changed ones would not work or be updated with `yarn serve`. Not sure how to fix this because the code only exists on your machine, and is not accessible to remix.

dwightjl:
> Not sure how to fix this because the code only exists on your machine, and is not accessible to remix.

Let's keep this as it is for now. It's better to have localhost ending up in URLs for the people who are testing things locally than it is to have it pull samples from docs.chain.link unexpectedly and wonder why samples are not the same as their local environment.

dwightjl:
Holding on merging this until after we merge #253.

dwightjl (changes requested):
Hold on this PR until we resolve issue #320.

dwightjl (approved):

StephenFluin:
@dwightjl I just realized we'll likely need to fix the headers on the netlify preview side of things. Have you verified that this works?

dwightjl:
I rebased another PR to pick up this change and indeed Remix is having trouble with the headers on Netlify. https://deploy-preview-339--dreamy-villani-0e9e5c.netlify.app/docs/intermediates-tutorial/ Older PRs still point to GitHub. The public site is functioning normally.

I'm not seeing where to update headers on Netlify, but I'll keep looking.

PR #331, opened 2021-08-25
add solana feeds

dmitrydao:


_tools/process-data.ts
```diff
@@ -93,19 +94,31 @@ for (let page of targetData) {
...
+      for (let contractKey of Object.keys(contents.contracts)) {
+        const contract = contents.contracts[contractKey];
         proxyList.push({
-          pair: proxy.name,
-          deviationThreshold: liveContracts[proxy.aggregator].deviationThreshold,
-          heartbeat: liveContracts[proxy.aggregator].heartbeat,
-          decimals: liveContracts[proxy.aggregator].decimals,
-          proxy: proxyKey,
+          pair: contract.name,
+          deviationThreshold: liveContracts[contractKey].deviationThreshold,
```
dwightjl, line 115:
```suggestion
            deviationThreshold: 'N/A',
```
This ends up being `null`. Can we override it with "N/A"?

_tools/networks.ts
```diff
@@ -132,5 +132,16 @@ export const NETWORKS = [
...
       },
     ],
+  },
+  {
+    page: "solana-price-feeds",
+    title: "Solana Price Feeds",
+    networks: [
+      {
+        name: "Solana Devnet",
+        url: "https://explorer.solana.com/address/",
```
dwightjl, line 142:
```suggestion
        url: "https://explorer.solana.com/?cluster=devnet",
```
Use the Devnet specific URL.

_includes/feed.liquid
```diff
@@ -3,7 +3,11 @@ layout: ./nodes.liquid
 section: smartContract
 ---
 
-<p>For implementation instructions, please refer to <a href="../get-the-latest-price/">Get the Latest Price</a>.</p>
+{% if metadata.solanafeeds %}
+    <p>For implementation instructions, please refer to <a href="/">Solana integration sample</a>.</p>
```
dwightjl, line 7:
```suggestion
    <p>The program address for this feed is publicly available in the <a href="https://explorer.solana.com/address/2yqG9bzKHD59MxD9q7ExLvnDhNycB3wkvKXFQSpBoiaE?cluster=devnet">Solana Devnet Explorer</a>.</p>
```

I will add a link to the implementation instructions later. For now, let's include the link to the program instead.

dwightjl (changes requested):

_tools/networks.ts
dmitrydao: The problem with this is that it will not open an address in the explorer. So we have an option of either pointing to the devnet cluster without an address or to the default explorer link (mainnet) with an address.

I thought it would be weird if you click on the address and it opens a blank explorer for you so I picked the second option. Thoughts?

_includes/feed.liquid
dmitrydao: Got it. Ok

_tools/process-data.ts
dmitrydao: `deviationThreshold` is a number so I tried to replace it with NaN but it returns null in any case. In the table, it shows up as N/A which is fine.

_tools/networks.ts
dwightjl: I tried this as a fix, but it's not a very elegant solution. It does work. https://github.com/dwightjl/documentation/commit/9d36974d5951f4ba121f9164b94b6f38291b5f7c

dwightjl (approved):
Excellent!

PR #333, opened 2021-08-25
Add basic support for Solana Devnet in our documentation.

dwightjl:
- Add a Solana network examples section with links to the GitHub Solana demo.
- Update the URLs to direct customers to the Solana Network Examples section.

Next steps will be to pull the documentation out of GitHub into a full set of documentation for the Solana demo with code sample explanations.

### DIFF
--- a/_includes/feed.liquid
+++ b/_includes/feed.liquid
@@ -4,12 +4,12 @@ section: smartContract
 ---
 
 {% if metadata.solanafeeds %}
-    <p>The program address for this feed is publicly available in the <a href="https://explorer.solana.com/address/2yqG9bzKHD59MxD9q7ExLvnDhNycB3wkvKXFQSpBoiaE?cluster=devnet">Solana Devnet Explorer</a>.</p>
+    <p>To learn how to implement these feeds, see the <a href="../get-the-latest-price/#solana-network-examples">Solana Examples for Consuming Price Feeds</a>.</p>
+    <p>To learn how to obtain SOL tokens on the Solana Devnet, see the <a href="https://docs.solana.com/cli/transfer-tokens#airdrop-some-tokens-to-get-started">Solana Documentation</a>.</p>
 {% else %}
-    <p>For implementation instructions, please refer to <a href="../get-the-latest-price/">Get the Latest Price</a>.</p>
+    <p>To learn how to implement these feeds, see the <a href="../get-the-latest-price/#ethereum-network-examples">Ethereum Examples for Consuming Price Feeds</a>.</p>
+    <p>For LINK token and Faucet details, see the <a href="../link-token-contracts/">LINK Token Contracts</a> page.</p>
 {% endif %}
-
-<p>For LINK token and Faucet details, please refer to <a href="../link-token-contracts/">LINK Token Contracts</a>.</p>
 
 <p>Note, off-chain equity and ETF assets are only traded during standard market hours (9:30 am - 4 pm ET M-F). Using these feeds outside of those windows is not recommended.</p>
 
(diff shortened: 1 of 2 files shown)


_includes/feed.liquid
```diff
@@ -4,13 +4,14 @@ section: smartContract
 ---
 
 {% if metadata.solanafeeds %}
-    <p>The program address for this feed is publicly available in the <a href="https://explorer.solana.com/address/2yqG9bzKHD59MxD9q7ExLvnDhNycB3wkvKXFQSpBoiaE?cluster=devnet">Solana Devnet Explorer</a>.</p>
+    <p>To learn how to implement these feeds, see the <a href="../get-the-latest-price/#solana-network-examples">Solana Examples for Consuming Price Feeds</a>.</p>
```
dmitrydao, line 7: Let's keep the same wording for all of our feeds.
`To learn how to implement these feeds, see..` -> `For implementation instructions, see..`

dmitrydao (commented):
LGTM

_includes/feed.liquid
dwightjl: Good call. It's best to be consistent. I still recommend that we re-word the sentence. Does that work, or do you prefer the existing sentence?

docs/Using Price Feeds/get-the-latest-price.html
```diff
@@ -87,6 +85,24 @@
     </div>
 </div>
 
+## Solana network examples
+
+Chainlink Price Feeds run live on the [Solana Devnet](https://explorer.solana.com/?cluster=testnet) with sub-second updates. Chainlink’s Solana deployment has no dependencies on external blockchain networks such as Ethereum.
+
+The program address for this price data feed is `2yqG9bzKHD59MxD9q7ExLvnDhNycB3wkvKXFQSpBoiaE` on the Solana Devnet. You can view the program in the [Solana Devnet Explorer](https://explorer.solana.com/address/2yqG9bzKHD59MxD9q7ExLvnDhNycB3wkvKXFQSpBoiaE?cluster=devnet).
```
PatrickAlphaC, line 92: Where did you get this program ID from?

PatrickAlphaC (commented):
Just curious on the program id, my demo shows people how to do both:

1. Deploy their own program
2. Interact with a deployed program

We will have to make it clear what we want them to do.

docs/Using Price Feeds/get-the-latest-price.html
PatrickAlphaC: I think ideally we will say something like:
```
This demo will show you how to:
1. Deploy your own program that uses Chainlink Data Feeds
2. Interact with an account that stores state from a deployed program

Please note, that for a solidity comparison, the combination of an account + program is equivalent to a smart contract on an EVM chain. The demo will show you how to work with a program you deploy, but you can refactor the client section to work with a program Id of your choice, as state and logic are separated in Solana. 
```

dwightjl: @mccordnate recommended that we highlight that program ID. It's also the owner ID for all the Devnet feeds: https://docs.chain.link/docs/solana-price-feeds/

For example, here is the BTC / USD feed:
https://explorer.solana.com/address/6dbkV6QCToTk6DRfuJyrGuz18kZ4rPUSHLLLVrryWdUC?cluster=devnet

mccordnate: Yeah, given the way Solana works, the demo is more important for how to actually use the price feeds than the actual program ID. I think it's just useful context for users to check against, though.

Thinking on this again in the morning, it's probably fine to leave it with the direct explorer link to the program.

dwightjl: I clarified these points and included the primer on what to expect from the demo.

PR #347, opened 2021-09-01
Add the Kovan Testnet to the list of Optimism feeds.

dwightjl:
- Add Ethereum Kovan to networks.ts to include the Kovan Testnet to the list of Optimism feeds.

The Netlify stating link won't show the change because it hasn't regenerated the addresses.json file on the server side, but I tested it locally.


_tools/networks.ts
```diff
@@ -153,6 +153,11 @@ export const NETWORKS = [
         url: "https://optimistic.etherscan.io/address/%s",
         source: "directory-ethereum-mainnet-optimism-1.json",
       },
+      {
+        name: "Ethereum Kovan",
+        url: "https://kovan-optimistic.etherscan.io/address/%s",
```
dmitrydao, line 158: I think this should be Optimism Kovan, at least that's how we name all of the other chains

dwightjl: Done.

dmitrydao:
LGTM. One question, do we have any faucet for Kovan Optimism that we can add here?

dmitrydao (approved):

mccordnate (approved):

PR #348, opened 2021-09-01
Add Netlify CORS headers for code samples

dwightjl:


dwightjl:
@StephenFluin
This seems to work. Do I need to add anything else to config.toml, or will it work in harmony with other Netlify config settings not defined in this file?

StephenFluin (approved):
Tested and seems to work, although I don't see the headers when I grab the file directly. Approving!

One small improvement would be to put at the top of this file why we have it, and that it only affects previews.

PR #354, opened 2021-09-03
redirect reference url

### LINKED ISSUE #353: https://docs.chain.link/reference results in Page Not Found (404)
Visiting page (link found via wiki : Node Operator Resources) results in 404. As far as I can tell, API documentation is kind of sparse so a working reference is pretty valuable.

dmitrydao:
fixes #353

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -96,6 +96,16 @@
         "source": "/docs/example-walkthrough",
         "destination": "/docs/intermediates-tutorial/",
         "type": 301
+      },
+      {
+        "source": "/docs/reference/",
+        "destination": "/",
+        "type": 301
+      },
+      {
+        "source": "/docs/reference",
+        "destination": "/",
+        "type": 301
       }
     ]
   }


dwightjl:
We have several references. Not just the price feed API reference. Instead of a redirect, can we create a reference list?

https://docs.chain.link/docs/chainlink-framework/
https://docs.chain.link/docs/chainlink-vrf-api-reference/
https://docs.chain.link/docs/price-feeds-api-reference/
https://docs.chain.link/docs/chainlink-node-api-reference/

dmitrydao:
> We have several references. Not just the price feed API reference. Instead of a redirect, can we create a reference list?
> 
> https://docs.chain.link/docs/chainlink-framework/
> https://docs.chain.link/docs/chainlink-vrf-api-reference/
> https://docs.chain.link/docs/price-feeds-api-reference/
> https://docs.chain.link/docs/chainlink-node-api-reference/

I think it would be confusing if we have too many API reference pages. Maybe we redirect to the main page instead?

dwightjl:
> I think it would be confusing if we have too many API reference pages. Maybe we redirect to the main page instead?

Yes for now that works.

dwightjl (approved):

PR #356, opened 2021-09-07
comment out unused vars from PriceConsumerV3 and RandomNumberConsumer

jmank88:
`PriceConsumerV3.sol` and `RandomNumberConsumer.sol` contain unused variables which ~prevents compiling and deploying as-is~ produce compiler warnings:
Screenshot from 2021-09-07 07-41-35
This change comments them out to ~allow for compiling without modification~ suppress the warning while keeping the type and name info for reference.

### DIFF
--- a/_includes/samples/PriceFeeds/PriceConsumerV3.sol
+++ b/_includes/samples/PriceFeeds/PriceConsumerV3.sol
@@ -21,11 +21,11 @@ contract PriceConsumerV3 {
      */
     function getLatestPrice() public view returns (int) {
         (
-            uint80 roundID, 
+            /* uint80 roundID */,
             int price,
-            uint startedAt,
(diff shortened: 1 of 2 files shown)


PatrickAlphaC:
This will deploy as-is. Those are just compiler warnings you can ignore, they will still let you deploy. We like to have them in the docs so you can see what all the variables are responding with.

jmank88:
My mistake on warning vs. error. Although, I wonder if this would still improve the UX. Notice that the keeper example uses this technique: https://docs.chain.link/docs/chainlink-keepers/compatible-contracts/#example-contract

dwightjl:
Let's hold off on this for now. Having the variables available to call is useful for educating developers even if the variables are not used in the sample application itself.

PR #361, opened 2021-09-08
show how to send an address in a request

### LINKED ISSUE #255: Show example of how to send an address in a request
The API reference here: https://docs.chain.link/docs/chainlink-framework/
Should be updated with an example of passing in an address in the request.

It's currently not clear how to do this. Passing an address as a uint will not work: https://github.com/smartcontractkit/chainlink/issues/4064

Talked to Steve about this, and the recommended way should be something along the lines of:

```javascript
req.addBytes('someAddress', abi.encodePacked(address(this)));
```

We should show this in the docs and make it the recommended way to do this. The only caveat here is that the value received on the CL node/external adapter will be a full word with leading zeroes, encoded with base64.

dmitrydao:
fix #255

This is probably the wrong implementation but it will be easier to solve it in the PR than going back and forth in the issue.

### DIFF
--- a/docs/Using Any API/api-reference.md
+++ b/docs/Using Any API/api-reference.md
@@ -963,12 +963,12 @@ function addBytes(
 Add a CBOR bytes type value to the run request parameters.
 
 ```javascript example
-function requestEmojiPopularity(bytes _unicode)
+function requestEmojiPopularity(address _address)
   public
 {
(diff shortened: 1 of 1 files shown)


dwightjl:
Closing this for now. The existing unicode example is still useful, so let's keep that and add a separate example for how to send an address.

PR #362, opened 2021-09-08
move all actions to shas

### LINKED ISSUE #71: Switch workflow action dependencies to shas
Use shas instead of tags or branches to improve security

StephenFluin:
This increases security by not allowing the code we execute to change.
Action devs can change tags, but they can't change the code for a sha.

Fixes #71


dwightjl (approved):
Smart! I like it.

PR #363, opened 2021-09-08
Fix element alignment on the docs homepage.

### LINKED ISSUE #188: Homepage div alignment is off
The centring of components seems to be slightly off on the home page.

Screenshot 2021-07-01 at 12 21 03

dwightjl:
Fixes #188


StephenFluin (approved):
I wonder why it was -8. I'm sure there was a reason at some point.

Oh well, Just checked at various widths and LGTM!

Issue #308, opened 2021-08-18
Missing environment variables


ArseniiPetrovich:
We have few nodes running on our end. We've been doing some code refactor recently and noticed that there are some env variables in our node that does not exist in the documentation. Here is the full list:
```
MIN_OUTGOING_CONFIRMATIONS
ETH_BALANCE_MONITOR_BLOCK_DELAY
ENABLE_BULLETPROOF_TX_MANAGER
OCR_TRACE_LOGGING
OCR_KEY_BUNDLE_ID
OCR_TRANSMITTER_ADDRESS
```
Can we have the documentation updated to reflect these variables?

P.S. I know, some of these variables are explained in other OCR docs, but it would be great to keep things in the single place.

dmitrydao:
@samsondav can you check these?

aelmanaa:
environment variables listed in https://docs.chain.link/chainlink-nodes/v1/configuration

PR #390, opened 2021-09-23
Update config vars to 0.10.14

samsondav:


dwightjl:
@samsondav Do we want to also include variables mentioned in #308 ?

dwightjl (approved):

PR #391, opened 2021-09-27
Add warning about single source providers to providers page.

danielgruesso:


docs/Data Provider Nodes/data-provider-nodes.md
```diff
@@ -11,6 +11,10 @@ Chainlink has facilitated the launch of several new oracle data services that al
 
 Data provider nodes allow customizable, on-demand API calls and the ability to deliver different data points on each request. This enables use cases where each requestor requires a different piece of data.
 
+> 🚧 Using single provider data nodes on production networks
+> 
+> When running an application on production networks, it is recommended to use nodes that make use of multiple data providers. Using a single source provider runs the risk of using incorrect data should the provider experience an issue.
```
dwightjl:
```suggestion
> When you run production applications on Mainnet, use nodes that have multiple data providers. If you use a node with only a single source provider and that provider experiences an issue, your application might receive incorrect or unreliable data.
```

dwightjl: `Using a single source provider`

Can we help customers quickly identify which providers have multiple data sources?

dwightjl: `nodes that make use of multiple data providers`

Use multiple nodes, use nodes with multiple data providers, or both? What exactly do we mean by nodes with multiple data providers? Nodes with multiple failover databases as backups?

dwightjl (changes requested):

docs/Data Provider Nodes/data-provider-nodes.md
danielgruesso: This has been updated, we shouldn't use `mainnet` here since it could be running on chains other than ethereum.

danielgruesso: I've updated this verbiage since it's not always possible to use multi-source providers.

danielgruesso: We're already doing this on market.link. Currently, none do. We're changing the UI there to reflect this.

docs/Data Provider Nodes/data-provider-nodes.md
```diff
@@ -11,6 +11,10 @@ Chainlink has facilitated the launch of several new oracle data services that al
 
 Data provider nodes allow customizable, on-demand API calls and the ability to deliver different data points on each request. This enables use cases where each requestor requires a different piece of data.
 
+> 🚧 Using single provider data nodes on production networks
+> 
+> When running a production application, be aware of the risks and implications of using a single data source for production networks; any error or omission in the provider's data may cause a potential negative impact on your application and its users.
```
dwightjl, line 16:
```suggestion
> When you run a production application, be aware of the risks and implications of using a single data source for production networks. Any error or omission in the provider's data might negatively impact your application and its users.
```

dwightjl (changes requested):

docs/Data Provider Nodes/data-provider-nodes.md
dwightjl: After the UI updates, let's a sentence here to tell users where to look for that information.

dwightjl (approved):

PatrickAlphaC (approved):

PR #406, opened 2021-09-30
Updated Price Feed conversion to be full contract example.  Incl. SafeMath warning note

melchermaxwell:
initial commit

initial commit

Added conversion formula.  .DS_Store to gitignore

Fixed typo and added using SafeMath

Edited the example to be in BTC not ETH

Updated conversion equation image path

Update _data/navigation.js

Co-authored-by: Dwight Lyle <dwightjl@gmail.com>

Update docs/Using Price Feeds/data-feeds-best-practices.md

Co-authored-by: Dwight Lyle <dwightjl@gmail.com>

Rm Data Feeds Best Practice page and moved text

typo

### DIFF
--- a/docs/Using Price Feeds/get-the-latest-price.html
+++ b/docs/Using Price Feeds/get-the-latest-price.html
@@ -91,19 +91,65 @@ This example uses [Web3.py](https://web3py.readthedocs.io/en/stable/) to retriev
 
 ## Getting a different price denomination
 
-Chainlink Data Feeds for asset prices can be used in combination to derive denominated price pairs in other currencies.
+Chainlink Price Feeds can be used in combination to derive denominated price pairs in other currencies.
 
-If you require a denomination other than what is provided, you can use two feeds to derive the pair that you need. For example, if you needed an BTC / EUR price, you could take the BTC / USD feed and the EUR / USD feed and derive BTC / EUR using [OpenZeppelin's SafeMath](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/SafeMath.sol) division.
+If you require a denomination other than what is provided, you can use two price feeds to derive the pair that you need. For example, if you needed a BTC / EUR price, you could take the BTC / USD feed and the EUR / USD feed and derive BTC / EUR using division.
 
 ![Request Model Diagram](/images/price-feed-conversion-equation.gif)
 
+> 🚧 Important
+> 
+> If you are using a Solidity version before 0.8.0 use [OpenZeppelin's SafeMath](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/SafeMath.sol) for division.
+
 
 ```solidity
-using SafeMath for uint256;
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.7;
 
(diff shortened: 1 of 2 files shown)


PatrickAlphaC:
Much better! One or two points.

docs/Using Price Feeds/get-the-latest-price.html
```diff
@@ -91,19 +91,65 @@
...
+     * Address: 0x6135b13325bfC4B00278B4abC5e20bbce2D6580e
+     * Decimals: 8
+     * Aggregator: EUR/USD
+     * Address: 0x0c15Ab9A0DB086e062194c273CC79f41597Bbf13
+     * Decimals: 8
+     */
+    constructor() {
+        priceFeedBtcUsd = AggregatorV3Interface(0x6135b13325bfC4B00278B4abC5e20bbce2D6580e);
+        priceFeedEurUsd = AggregatorV3Interface(0x0c15Ab9A0DB086e062194c273CC79f41597Bbf13);
+        decimals = 10**8;
```
PatrickAlphaC, line 129: Let's not hard code decimals like this. You can actually call the `decimals` function to get the number of decimals. ie:

```
priceFeedEurUsd.decimals()
```

melchermaxwell: There are two situations where this is conversion is used:
(1) USD-base denominated (illustrated in above)
(2) blockchain base currency denominated (BNB, ETH, etc.) 

Most all conversions will be through USD denominations on both pairs which will be 8 decimals.  If it is not through a USD denominated base then you would be using multiplication.
Ex. ALPHA/BNB * BNB/USD = ALPHA/USD
ALPHA/BNB (18 decimals)
BNB/USD (8 decimals)
.002257198340965213 * 418.01476442
2257198340965213 * 41801476442
You would need to divide the result by the sum of the two decimals.  Regardless, the example in the docs should illustrate Option 1 as we would be more willing to spin up a price feed for a requested USD denomination if needed.  As for not hard coding, that's fine but should I just get the decimals of the numerator and use that?

PatrickAlphaC: Chatted with Max. Going to have 2 functions to address this.

PatrickAlphaC:
We are starting to look good!

Only one other comment, it's highly likely that users will want the WEI version of these (ie, with 18 decimals places). 

Also, I made the suggestions a lot smaller with some solidity tricks. Let me know if you @melchermaxwell find those more confusing, or make it easier to read. If it's more confusing, we can put it back to the way you had it.

PatrickAlphaC:
> @PatrickAlphaC Accepted your suggestions. Looks good and appreciate the Solidity tips i.e. `( , int priceBtcUsd, , , ) = priceFeedBtcUsd.latestRoundData();` I ran both sets of code and remix and found an extra parenthesis and a variable moved from the wrong example. Fixed and both. They both ran correctly but please confirm. Other than that, let me know if you have any additional comments. Already rebased to main and squashed commits. :)

Whoops, thank you! Looks like something is breaking. Once those things are fixed, we can merge!

An upgrade we can do after that'll probably want to happen is to convert everything to having 18 decimals.

PatrickAlphaC:
Awesome!

just to note, if customers are asking for conversion to 18 decimals, we should add that next.

PR #420, opened 2021-10-05
Remove hard-coded Solidity versions from Remix links.

dwightjl:
- Remove hard-coded Solidity versions from Remix links. Allow Remix to automatically select the compiler version from `pragma`.
- Update the version numbers for https://github.com/smartcontractkit/chainlink/blob/master/contracts/ files where necessary.
- Auto-remove trailing whitespace.

### DIFF
--- a/_src/deploy-to-github.js
+++ b/_src/deploy-to-github.js
@@ -3,7 +3,7 @@ document.addEventListener('DOMContentLoaded', () => {
   const host = window.location.hostname;
   if (host !== 'docs.chain.link') {
     // Rewrite Remix URLs with current hostname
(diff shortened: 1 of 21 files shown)


ritika-b (approved):

PatrickAlphaC:
Actually... looks like you're right, it does auto-detect. That seems to be the easiest way to do it. I can't seem to replicate how I got the errors, so I'm assuming I was doing something wrong. 

Can we just keep tabs on this change? If we start running into issues where the version isn't automatically 100% correct, we can swap back to adding sol versions in the URL?

PatrickAlphaC (approved):

dwightjl:
I triple-checked the functionality in Remix. It won't compile a contract outside of the specified `pragma` range. It auto-updates the compiler version to match the `pragma` even if you manually select an incompatible version. If you specify a range like `>=0.4.0 <0.6.0` it defaults to the latest acceptable version in the range unless you select another version within that acceptable range.

PR #423, opened 2021-10-06
Add notice about index price feeds based on liquid pairs

dwightjl:
Preview: https://deploy-preview-423--dreamy-villani-0e9e5c.netlify.app/docs/ethereum-addresses/


dwightjl:
Perhaps we could turn this into an info icon with a message that expands on hover. I'll look into it.

PR #452, opened 2021-10-27
Update links to external adapters

thodges-gh:
We get engineering candidates that use my old (and unmaintained) EA template where they should only be using our monorepo due to the docs pointing to them. I have no intention of updating the old templates so let's get them off the docs.


dwightjl (approved):

dwightjl:
Opened https://github.com/smartcontractkit/documentation/issues/465 to track the fork and update task for the previous examples.

PR #463, opened 2021-10-31
Mention FEATURE_EXTERNAL_INITIATORS

sudoFerraz:
There is no mention on the page or on the external initiator topic that it is necessary to enable the flag FEATURE_EXTERNAL_INITIATORS=true on the .env file of the node. I'm sure other people got stuck on this like me for quite some time.


dwightjl (approved):
Thank you for adding this. Let's move the notice up to the top of the page and add it to the introduction as well.

PR #464, opened 2021-11-01
added learning resources

PatrickAlphaC:


PatrickAlphaC:
Hi @ritika-b,

Now that Cryptozombies and Chainshot are starting or about to start supporting Chainlink content (yay!), I think its time we had a page that points out to showing people how to build smart contracts and hybrid smart contracts.

Right now we have #hackathon-resources page... but I feel like maybe the two of these pages should be merged into one? I'm not sure where this list should live, any suggestions?

PR #471, opened 2021-11-04
Add learn more footer section.

dwightjl:
- Make cards fully clickable on homepages
- Remove unused classes from homepage cards
- Add cta-learnmore section
- Add Stack Overflow icon


ZakAyesh:
The only concern I have here is "Talk to an expert" button label. This sends people to the same typeform found on chain.link, which is for businesses requesting help integrating and such. Coming from the docs though, developers may think this is a way to get 1-0n-1 technical help, instead of first heading to StackOverflow or Discord, and fill the type form with technical support questions.

How do we feel about labeling this button with something else, such as "Need integration support? Talk to an expert"?
@dwightjl @StephenFluin

dwightjl:
> How do we feel about labeling this button with something else, such as "Need integration support? Talk to an expert"? 

Done! Great suggestion. Confirmed the change with the design and growth team.

StephenFluin (approved):

PR #476, opened 2021-11-05
refactor keepers config for clarity and manychain

StephenFluin:


PatrickAlphaC (commented):
Looks ok, I can't speak much for some of the typescript stuff going on. 

However, it looks like polygon is blank for a lot of the new values?

dwightjl:
> However, it looks like polygon is blank for a lot of the new values?

Fixed with another push. Polygon is not ready to be published yet. It just needed to be removed from metadata.

dwightjl (approved):

PR #485, opened 2021-11-13
Add LINK address for Polygon Networks 

JimLynchCodes:


dwightjl (approved):
Let's remove these addresses and point people to the LINK Token Contracts page instead. Commit added.

PR #489, opened 2021-11-15
added info on gas lanes

tkaraivanov:


docs/Using Randomness/vrf-deployments.md
```diff
@@ -33,6 +33,9 @@ You can view them by running `getFeeConfig` on the coordinator.
...
 - `uint24 reqsForTier2`, `uint24 reqsForTier3`, `uint24 reqsForTier4`, `uint24 reqsForTier5` - tier boundaries in terms of requests made using a given subscription.
 
+# Gas prices
+To avoid costly responses during times of gas price volatility, there is a limit on the maximum gas price that VRF response transactions will be submitted with. Since every use case is different, there are several different VRF jobs set up on each blockchain with varying gas caps, each corresponding to a different keyHash. Contract developers can reference the appropriate keyHash according to how much they are willing to pay for gas depending on the specific use case. Note that these are maximum gas prices and will only be used if current network conditions require it - otherwise a more appropriate lower gas price will be used to submit the response. If your use case requires a different gas price cap from what is currently available, please [reach out](https://chainlinkcommunity.typeform.com/to/OYQO67EF) to our team.
```
dwightjl, line 37:
```suggestion
To avoid costly responses during gas price volatility, VRF response transactions are submitted with a limit on the maximum gas price. Because every use case is different, different VRF jobs are set up on each blockchain with different gas caps. Each VRF job corresponds to a different keyHash. Contract developers can reference the appropriate keyHash according to how much they are willing to pay for gas for each specific use case. Note that these are maximum gas prices and are used only if current network conditions require it. Otherwise, a more appropriate lower gas price is used to submit the response. If your use case requires a different gas price cap than what is currently available, [contact our team](https://chainlinkcommunity.typeform.com/to/OYQO67EF).
```

dwightjl:
Reworded several sentences.
Squashed and rebased against the add-vrf-v2-docs branch.
Merging with the add-vrf-v2-docs branch.

PR #529, opened 2021-12-01
match global footer CTA spacing & placement to mockup

StephenFluin:


_src/docs.css
```diff
@@ -1313,6 +1319,10 @@ input[type='reset'] {
     margin-right: auto;
     margin-left: auto;
   }
+
+  .cta-learnmore > * {
+    width: 100%;
```
dwightjl, line 1324:
```suggestion
    width: 370px;
```

At 100% the cards stretch out when the browser is less than 991px wide. Can we limit it, or do you think it looks better at 100%?

dwightjl (changes requested):

_src/docs.css
StephenFluin: Thre goal here is to jump to 100% with before they all wrap into one call. Where this happens will depend on the specific content in the boxes. Using the existing breakpoint gives us a little bit of buffer.

dwightjl: Sounds good.

dwightjl (approved):

PR #539, opened 2021-12-05
add-detailed-hw-reqs

rohhan:
Added details for minimum and recommended HW recs for Chainlink node and PostgreSQL DB.


docs/Node Operators/running-a-chainlink-node.md
```diff
@@ -16,10 +16,41 @@ It's important to note that nodes can fulfill requests for open APIs out-of-the-
...
+
+Your Chainlink node should be run on a server that has a public IP address.
+
+### Minimum
+
+To get started running a Chainlink node, you will need a machine with at least **2 cores** and **4 GB of RAM**. 
+
+### Recommended 
+
+The requirements for running a Chainlink node scale as the number of jobs your node services also scales.  For nodes with over 100 jobs, you will need at least **4 cores** and **8GB of RAM**.
```
dwightjl, line 33:
```
For nodes with over 100 jobs, you will need at least **4 cores** and **8GB of RAM**.  
```
Do we have a better metric for this like Job Runs per minute? I also assume the complexity of the job is a factor.

docs/Node Operators/running-a-chainlink-node.md
```diff
@@ -16,10 +16,41 @@ It's important to note that nodes can fulfill requests for open APIs out-of-the-
...
 
+# Hardware Requirements
+
+## Chainlink Node
+
+Your Chainlink node should be run on a server that has a public IP address.
+
+### Minimum
+
+To get started running a Chainlink node, you will need a machine with at least **2 cores** and **4 GB of RAM**.
```
dwightjl, line 29:
```suggestion
**2 cores**
```
Do we want to provide a cutoff for specific processor generations? For example, on the extreme end I wouldn't expect a [Core 2 Duo E6300](https://ark.intel.com/content/www/us/en/ark/products/27248/intel-core2-duo-processor-e6300-2m-cache-1-86-ghz-1066-mhz-fsb.html) would be sufficient.

I assume we don't support ARM either. Correct?

docs/Node Operators/running-a-chainlink-node.md
```diff
@@ -16,10 +16,41 @@ It's important to note that nodes can fulfill requests for open APIs out-of-the-
...
+
+### Minimum
+
+The minimum requirements for the database are **2 cores**, **4GB of RAM**, and **100 GB of storage**.
+
+### Recommended
+
+Similar to the Chainlink node, requirements increase as you service more jobs.  For more than 100 jobs, your database server will need at least **4 cores**, **16 GB of RAM**, and **100 GB of storage**. 
+
+If using AWS, use dedicated IOPS (not the burstable one).
```
dwightjl, line 48:
```suggestion
If you run your node on AWS, use an instance type with dedicated core time. [Burstable Performance Instances](https://aws.amazon.com/ec2/instance-types/#Burstable_Performance_Instances) have a limited number of [CPU credits](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-credits-baseline-concepts.html).
```

dwightjl (changes requested):

docs/Node Operators/running-a-chainlink-node.md
rohhan: This is the most detail we have available at the moment.  We can look into making things more specific in the future, but right now we don't have that info.

docs/Node Operators/running-a-chainlink-node.md
rohhan: I like your improvement here.  Is it possible to make the second sentence a bit more explicit?  "Burstable Performance instances have...and so should not be used"

docs/Node Operators/running-a-chainlink-node.md
rohhan: Just checked with the team and we should be okay without specifying the additional details mentioned.
With regards to the first point, we don't want to be too specific since we will need to keep updating if so.
With regards to the second point, as far as we know, it is supported, though we may need to do some testing in the future.

docs/Node Operators/running-a-chainlink-node.md
dwightjl: Made an edit to the suggestion. Is that more clear?

dwightjl (approved):

PR #560, opened 2021-12-19
updateCron

codetributor:
Another person was wondering why the cron job did not have a oracle contract address. In the Chainlink disccord they were wondering how they get paid in cron jobs. I'm still not sure the best way to setup up payment for cron jobs, but I updated some explaination to the cron docs. I sought clarification on stackoverflow and got answers from a chainlink member. I pretty much copied and pasted to update the cron job. I hope I'm helping, let me know if these edits are acceptable.


docs/Oracle Jobs/Job Types/cron.md
```diff
@@ -6,6 +6,7 @@ permalink: "docs/jobs/types/cron/"
 ---
 
 Executes a job on a schedule. Does not rely on any kind of external trigger.
+This means it's triggered based on some condition that the Chainlink node evaluates, and not triggered externally via a smart contract. Because of this, the node isn't paid in LINK tokens for processing the request like it does for API calls, because it's initiating a request itself as opposed to receiving a request (and payment) from on-chain.
```
ritika-b, line 9:
```suggestion
This it's triggered based on some condition that the Chainlink node evaluates, and not triggered externally via a smart contract. Because of this, the node isn't paid in LINK tokens for processing the request like it does for API calls, because it's initiating a request itself as opposed to receiving a request (and payment) from on-chain.
```

PR #535, opened 2021-12-02
Update functions for 0.3.0 contracts.

dwightjl:

PR #571, opened 2021-12-23
requestOracleData should be sendChainlinkRequest

D-Roch:
ChainlinkClient functions requestOracleData and requestOracleDataFrom have been changed to sendChainlinkRequest and sendChainlinkRequestTo respectively.


dwightjl:
Duplicate with #535

PR #585, opened 2022-01-15
Replace fulfillRandomness external with internal function type

DSergiu:
Docs [here](https://docs.chain.link/docs/chainlink-vrf-api-reference/#requestrandomness) mention `external` function type for `fulfillRandomness`.

image

If this function definition is used, it could ofc have security implications allowing an external malicious call with pre-cooked input.


dwightjl (approved):
This is correct. Thank you for the PR!

PR #604, opened 2022-01-31
[1.1.1] Documentation changes for 1.1.1

samsondav:


docs/Node Operators/configuration-variables.md
```diff
@@ -208,6 +214,12 @@ This variable enables some chain-specific hacks and optimizations. It is recomme
 
 Setting `CHAINLINK_DEV` to `true` enables development mode. This setting is not recommended for production deployments. It can be useful for enabling experimental features and collecting debug information.
 
+### SHUTDOWN_GRACE_PERIOD
+
+- Default: `"5s"`
+
+When Chainlink node is shutting down gracefully and exceeded this grace period, it gives up trying to stop all (sub)services gracefully, closes DB connection and quits.
```
jmank88, line 221: Should we say a little more about why and how to calibrate this?
```suggestion
If the grace period is exceeded during shutdown, the node will (1) abandon blocking (sub)services, (2) release the DB lock and connection, and (3) exit immediately. This should be set _less_ than your systems `SIGKILL` timeout.
```

pinebit: Thank you @jmank88 !

PR #610, opened 2022-02-03
Update logging levels

samsondav:


pinebit (approved):

prashantkumar1982 (approved):

docs/Node Operators/configuration-variables.md
```diff
@@ -408,11 +408,13 @@ By default, Chainlink nodes write log data to `$ROOT/log.jsonl`. The log directo
 The `LOG_LEVEL` environment variable determines both what is printed on the screen and what is written to the log file.
 
 The available options are:
+- `"trace"`: Used only in development, off by default and not accessible on production builds.
```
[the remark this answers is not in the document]
dwightjl: Let's leave it in. If its in the code, we should document it here. Node operators documentation is not just for customers. This logging level might be relevant only to a narrow audience but development might read this page and benefit from the full list of options. https://github.com/smartcontractkit/chainlink/blob/4f776ea86a15c3f23400fa60c4c2f1985aeb750f/core/logger/logger.go#L39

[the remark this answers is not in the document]
dwightjl: Let's leave this in. If its in the code, this documentation should match it. Devs might use this page and benefit from the full list of options. Trace seems to be useful for a specific audience: https://github.com/smartcontractkit/chainlink/blob/4f776ea86a15c3f23400fa60c4c2f1985aeb750f/core/logger/logger.go#L39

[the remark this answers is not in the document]
jmank88: Trace is not just another log level, so this is incorrect as currently stated. It is not an option for `LOG_LEVEL`, and will not be emitted as a separate "level" either.

docs/Node Operators/configuration-variables.md
```diff
@@ -408,11 +408,13 @@ By default, Chainlink nodes write log data to `$ROOT/log.jsonl`. The log directo
...
-- `"error"`: Something bad happened. Node operators might need to take urgent action based on this error.
-- `"panic"`: Something very bad happened. Node operators should take immediate action to fix this.
+- `"warn"`: A mild error happened that might require non-urgent action. Node operators should check these once in a while to see whether anything stands out. These can happen due to factors outside of the control of the node operator (e.g. unexpected responded from a remote API, spurious networking errors etc).
+- `"error"`: Something bad happened. The node might be running in a partially broken state. Node operators might need to take action to remedy this error (e.g. deprecation warnings, misconfiguration causing a job to fail to run etc).
+- `"crit"`: A critical error occured. The node might be fully broken. Node operators should take immediate action to fix this (e.g. node was not able to boot because a network socket could not be opened, database became inaccessible etc).
+- `"panic"`:An unrecoverable problem was encountered that was unexpected. Node operators should try to restart their nodes, and notify Chainlink team of a bug.
```
dwightjl, line 417:
```suggestion
- `"warn"`: A mild error occurred that might require non-urgent action. Check these warnings semi-regularly to see if any of them require attention. These warnings usually happen due to factors outside of the control of the node operator. Examples: Unexpected responded from a remote API or inaccurate networking errors.
- `"error"`: An error occurred that might have put the node in a partially broken state. Node operators might need to take action to remedy this error. Examples: Deprecation warnings or incorrectly configured settings that cause a job to fail.
- `"crit"`: A critical error occurred. The node might be completely stopped and non-responsive. Node operators should take immediate action to fix these errors. Examples: The node was not able to boot because a network socket could not be opened or the database became inaccessible.
- `"panic"`: An unrecoverable error occurred that was unexpected. Node operators should try to restart their nodes, and notify the Chainlink team of a bug.
```

docs/Node Operators/configuration-variables.md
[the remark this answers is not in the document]
samsondav: I suggest we remove it from these docs. These are public facing - it's ok for us to have private internal developer-only docs that are different.

Not all env variables are documented for this reason.

docs/Node Operators/configuration-variables.md
samsondav: @dwightjl `inaccurate networking errors` <- not really sure what this means

docs/Node Operators/configuration-variables.md
```diff
@@ -408,11 +408,13 @@ By default, Chainlink nodes write log data to `$ROOT/log.jsonl`. The log directo
...
-- `"error"`: Something bad happened. Node operators might need to take urgent action based on this error.
-- `"panic"`: Something very bad happened. Node operators should take immediate action to fix this.
+- `"warn"`: A mild error occurred that might require non-urgent action. Check these warnings semi-regularly to see if any of them require attention. These warnings usually happen due to factors outside of the control of the node operator. Examples: Unexpected responded from a remote API or inaccurate networking errors.
+- `"error"`: An error occurred that might have put the node in a partially broken state. Node operators might need to take action to remedy this error. Examples: Deprecation warnings or incorrectly configured settings that cause a job to fail.
+- `"crit"`: A critical error occurred. The node might be completely stopped and non-responsive. Node operators should take immediate action to fix these errors. Examples: The node was not able to boot because a network socket could not be opened or the database became inaccessible.
```
kaleofduty, line 416:
```suggestion
- `"crit"`: A critical error occurred. The node might be unable to function. Node operators should take immediate action to fix these errors. Examples: The node was not able to boot because a network socket could not be opened or the database became inaccessible.
```

docs/Node Operators/configuration-variables.md
```diff
@@ -408,11 +408,13 @@ By default, Chainlink nodes write log data to `$ROOT/log.jsonl`. The log directo
...
 
 The available options are:
+- `"trace"`: Used only in development, off by default and not accessible on production builds.
 - `"debug"`: Useful for forensic debugging of issues.
 - `"info"`: High level informational messages.
-- `"warn"`: Something unexpected happened that might need non-urgent action. Node operators should check these once in a while to see whether anything stands out (e.g. deprecation warnings).
-- `"error"`: Something bad happened. Node operators might need to take urgent action based on this error.
-- `"panic"`: Something very bad happened. Node operators should take immediate action to fix this.
+- `"warn"`: A mild error occurred that might require non-urgent action. Check these warnings semi-regularly to see if any of them require attention. These warnings usually happen due to factors outside of the control of the node operator. Examples: Unexpected responded from a remote API or inaccurate networking errors.
+- `"error"`: An error occurred that might have put the node in a partially broken state. Node operators might need to take action to remedy this error. Examples: Deprecation warnings or incorrectly configured settings that cause a job to fail.
```
kaleofduty, line 415:
```suggestion
- `"error"`: An error occurred that is not expected during regular operation of a well-maintained node. Node operators might need to take action to remedy this error. Check these regularly to see if any of them require attention. Examples: Use of deprecated configuration options or incorrectly configured settings that cause a job to fail.```

File: docs/Node Operators/configuration-variables.md [hunk 5]
```diff
@@ -408,11 +408,13 @@ By default, Chainlink nodes write log data to `$ROOT/log.jsonl`. The log directo
 The `LOG_LEVEL` environment variable determines both what is printed on the screen and what is written to the log file.
 
 The available options are:
+- `"trace"`: Used only in development, off by default and not accessible on production builds.
 - `"debug"`: Useful for forensic debugging of issues.
 - `"info"`: High level informational messages.
-- `"warn"`: Something unexpected happened that might need non-urgent action. Node operators should check these once in a while to see whether anything stands out (e.g. deprecation warnings).
-- `"error"`: Something bad happened. Node operators might need to take urgent action based on this error.
-- `"panic"`: Something very bad happened. Node operators should take immediate action to fix this.
+- `"warn"`: A mild error occurred that might require non-urgent action. Check these warnings semi-regularly to see if any of them require attention. These warnings usually happen due to factors outside of the control of the node operator. Examples: Unexpected responded from a remote API or inaccurate networking errors.
```
kaleofduty, line 414: What is an inaccurate networking error?

docs/Node Operators/configuration-variables.md
jmank88: > Examples: The node was not able to boot because a network socket could not be opened or the database became inaccessible.
Are these good examples? They sound like fatal problems to me.

docs/Node Operators/configuration-variables.md
```diff
@@ -408,11 +408,13 @@ By default, Chainlink nodes write log data to `$ROOT/log.jsonl`. The log directo
...
+- `"error"`: An error occurred that might have put the node in a partially broken state. Node operators might need to take action to remedy this error. Examples: Deprecation warnings or incorrectly configured settings that cause a job to fail.
+- `"crit"`: A critical error occurred. The node might be completely stopped and non-responsive. Node operators should take immediate action to fix these errors. Examples: The node was not able to boot because a network socket could not be opened or the database became inaccessible.
+- `"panic"`: An unrecoverable error occurred that was unexpected. Node operators should try to restart their nodes, and notify the Chainlink team of a bug.
```
jmank88, line 417: > Node operators should try to restart their nodes

Why? Panics are not necessarily fatal. A task can panic and be recovered and marked as an error, without affecting any other systems.

kaleofduty: I imagine that there are concerns about the panic having left some part of the system in an invalid state

jmank88: If that is truly a valid concern, then we should not be recovering and stopping the system from shutting down. It doesn't make sense to recover and keep the node alive while also expecting node ops to notice and manually restart.

docs/Node Operators/configuration-variables.md
kaleofduty: There definitely are cases where we cannot open a socket but the node overall might still function. 

IIUC fatal kills the node. But from within libocr we likely shouldn't kill the node, it might also be doing non OCR things.

jmank88: Is a node in this partial state considered 'booted', or 'not able to boot'? From a developer perspective I was thinking that 'not able to boot' literally meant a blocking error during startup which is effectively fatal, and that this example/language should be reserved for that description. I think that generally we should try to document the _least severe_ examples at each level to illustrate the boundary most clearly.

docs/Node Operators/configuration-variables.md
dwightjl: @samsondav What are some examples of spurious networking errors? Perhaps "misleading" is a better term to use?

docs/Node Operators/configuration-variables.md
dwightjl: What are some examples of spurious networking errors? Perhaps "misleading" is a better term to use?

docs/Node Operators/configuration-variables.md
dwightjl: I changed the term to "misleading".

docs/Node Operators/configuration-variables.md
dwightjl: I changed the term to "misleading".

docs/Node Operators/configuration-variables.md
dwightjl: Do we have a consensus on what the correct node operator response should be for `panic`? I changed it a little in my last commit:

```
If the node is unresponsive, node operators should try to restart their nodes and.....
```

jmank88: Yeah that sounds good.
I'm worried about 'unrecoverable' being too suggestive - What about:
> An exceptional error occurred that could not be handled.

dwightjl: Updated. Looks like this PR is good to go if there are no other concerns.

dwightjl (approved):

dwightjl:
@jmank88 & @samsondav is this good to merge?

jmank88 (approved):

PR #617, opened 2022-02-10
Comment unused variables to avoid warning messages

dwightjl:


_includes/samples/PriceFeeds/PriceConsumerV3.sol
```diff
@@ -20,12 +20,15 @@ contract PriceConsumerV3 {
      * Returns the latest price
      */
     function getLatestPrice() public view returns (int) {
-        (
-            uint80 roundID, 
+        (            
+            // The roundID, startedAt, timeStamp, and answeredInRound variables
+            // indicate when this specific price value was determined, which
+            // can be useful for comparing against historical price data.
```
PatrickAlphaC, line 26: I think you could probably leave this off, having the comments for where the other variables would be created I think is perfect.

dwightjl: Done.

aelmanaa (approved):

PR #632, opened 2022-02-21
Update get-a-random-number.md

aelmanaa:
_VRFConsumerBaseV2.sol_ [link](https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/dev/VRFConsumerBaseV2.sol) is not correct; Hence readers are redirected to this page:

image

the correct link is :
https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/VRFConsumerBaseV2.sol


dwightjl (approved):
Good catch. I pushed an additional commit to update the migration page.

PR #673, opened 2022-03-23
Updated example for LargeResponses with V2 JobSpec

melchermaxwell:
Updated the **Large Responses** JobSpec page with a V2 example.


docs/Example Job Specs/large-response-job.md
```diff
@@ -48,3 +48,38 @@ This is an example v1 job spec for returning multiple responses in 1 Chainlink A
   ]
 }
 ```
+
+This is an example v2 job spec for returning large responses in 1 Chainlink API Call. 
+
+
+```jpv2
```
dwightjl, line 55: Perhaps move this above the JSON example because TOML is the preferred spec.

dwightjl (changes requested):

dwightjl (approved):

PR #679, opened 2022-03-29
Added missing [mode] parameter.

pinebit:
`cborparse` task has parameter "mode" which was undocumented.


jmank88 (approved):

docs/Oracle Jobs/Task Types/task_cborparse.md
```diff
@@ -11,6 +11,7 @@ CBOR Parse tasks parse a CBOR payload, typically as part of a Direct Request wor
 **Parameters**
 
 - `data`: a byte array containing the CBOR payload.
+- `mode`: optional string: `diet` (default) or `standard`. In `diet` mode, the incoming CBOR is expected to be a map.
```
dwightjl, line 14:
```suggestion
- `data`: A byte array containing the CBOR payload.
- `mode`: An optional parameter that specifies how to parse the incoming CBOR. The default mode is `diet`, which expects the input to be a map. Set the mode to `standard` to pass literal values through "as-is". Empty inputs return nil.
```

Expanded the text based on the code comments here: https://github.com/smartcontractkit/chainlink/blob/d39dc15baddafb193b8919102bf8519c789f9083/core/cbor/cbor.go#L38

dwightjl (changes requested):

dwightjl (approved):

PR #706, opened 2022-04-15
e2e test apiconsumer

aelmanaa:
- create test for apiconsumer
- add github actions to e2e test apiconsumer on kovan


.github/workflows/test.yml
```diff
@@ -8,12 +8,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+        uses: actions/checkout@v3
```
dwightjl, line 11:
```suggestion
        uses: actions/checkout@dcd71f646680f2efd8db4afa5ad64fdcba30e748
```

I think Stephen had a reason to pin to SHA rather than version numbers. Also 3.0.1 [released](https://github.com/actions/checkout/releases) a few days ago.

package.json
```diff
@@ -54,5 +54,28 @@
     "trailingComma": "es5",
     "singleQuote": true,
     "printWidth": 120
+  },
+  "devDependencies": {
```
dwightjl, line 58: Are all of these necessary? The tests seem to work fine after I remove the following dependencies:

```
"@typescript-eslint/eslint-plugin": "^5.19.0",
"@typescript-eslint/parser": "^5.19.0",
"eslint": "^8.13.0",
"eslint-config-prettier": "^8.5.0",
"eslint-config-standard": "^16.0.3",
"eslint-plugin-import": "^2.26.0",
"eslint-plugin-node": "^11.1.0",
"eslint-plugin-prettier": "^4.0.0",
"eslint-plugin-promise": "^6.0.0",
"prettier": "^2.6.2",
"prettier-plugin-solidity": "^1.0.0-beta.19",
"solhint": "^3.3.7",
```

After removing these, I get only the following unmet peer dependencies:

```
warning " > @typechain/hardhat@6.0.0" has unmet peer dependency "@ethersproject/abi@^5.4.7".
warning " > @typechain/hardhat@6.0.0" has unmet peer dependency "@ethersproject/providers@^5.4.7".
warning " > @typechain/ethers-v5@10.0.0" has unmet peer dependency "@ethersproject/abi@^5.0.0".
warning " > @typechain/ethers-v5@10.0.0" has unmet peer dependency "@ethersproject/bytes@^5.0.0".
warning " > @typechain/ethers-v5@10.0.0" has unmet peer dependency "@ethersproject/providers@^5.0.0"
```

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  test-contracts:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+      - name: Install Dependencies
+        run: yarn
+      - name: Create .env
+        run: |
```
dwightjl, line 15: Can we remove this if GitHub is able to obtain the secrets from `${{ secrets.ETHERSCAN_API_KEY }}` etc. Is there a reason for GitHub to also create a .env file?

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  push:
+  schedule:
+    - cron: '0 20 * * */1'
+jobs:
+  test-contracts:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
```
dwightjl, line 11:
```suggestion
        uses: actions/checkout@dcd71f646680f2efd8db4afa5ad64fdcba30e748
```

package.json
```diff
@@ -54,5 +54,28 @@
...
+    "@nomiclabs/hardhat-etherscan": "^3.0.3",
+    "@typechain/ethers-v5": "^10.0.0",
+    "@typechain/hardhat": "^6.0.0",
+    "@types/chai": "^4.3.1",
```
dwightjl, line 62: Do we need both `"@types/chai": "^4.3.1",` and `"chai": "^4.3.6",`?

dwightjl (changes requested):

.github/workflows/test.yml
StephenFluin: Tags can be changed, which means it's an avenue for remote code execution (They push a new commit with some malicious code and then change the tag so we get the new version and execute it automatically). It's unlikely for GitHub to be compromised like that but it's definitely possible with other repositories, so it's safer to use a SHA for all workflows.

package.json
StephenFluin: @types/chai adds in the types for Chai so that any typescript you write can be type checked. If we're manually importing or using Chai, we should add this. If it's only included transitively we can skip.

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+          echo KOVAN_URL=$KOVAN_URL >> .env
+          echo MNEMONIC=$MNEMONIC >> .env
+        env:
+          ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+          KOVAN_URL: ${{ secrets.KOVAN_URL }}
+          MNEMONIC: ${{ secrets.MNEMONIC }}
+      - name: Compile contracts
+        run: npx hardhat compile
+      - name: Test API consumer
+        run: npx hardhat test test/any-api/apiconsumer.ts --network kovan
```
StephenFluin, line 27: Who is going to be monitoring the Kovan wallet to make sure this keeps working? How do we minimize the amount of extra effort here?

test/any-api/apiconsumer.ts
```diff
@@ -0,0 +1,74 @@
+import { ethers } from 'hardhat';
```
StephenFluin, line 1: Should the filename be apiconsumer.ts or api-consumer.ts? I think kebab case is preferred in TS.

StephenFluin (changes requested):

dwightjl:
Let's move this to another repo for end-to-end automated tests of the DevRel nodes.

PR #772, opened 2022-05-24
Remove external state variable visibility

bobanm:
Solidity supports only `private`, `internal` and `public` visibility for state variables.
Only functions can have `external` visibility.


aelmanaa:
Hello @bobanm 

[Indeed](https://docs.soliditylang.org/en/v0.8.14/contracts.html#state-variable-visibility). good catch, thanks a lot!

aelmanaa (approved):

PR #793, opened 2022-06-08
Added non-root user to chainlink deploy command.

Inbell1s:


aelmanaa:
Hi @Inbell1s 

Sorry for the delay! Actually, the recommended images in the docs are non root images. To check:

1.  Exec into a container

        `docker exec -it chainlink /bin/bash`

1.  Run `id`

          ```bash
          chainlink@bb012ed6b8d6:~$ id
          uid=14933(chainlink) gid=14933(chainlink) groups=14933(chainlink)
          ```
          
 As you can see, the user and its group are non-root

PR #844, opened 2022-07-05
Add new section for explanation and examples of using chainlinkVRF to Handle Different function logic request

franfran20:
## Description

In the documentation there isn't really any guide showing that different functions can make their own specific requests and the `fulfillRandomWords` function will execute a logic that is specific to the function that made that request.
I was personally working on a hackathon and initially thought maybe there was a way to have multiple `fulfillRandomWords` function but I doubt tha'ts the case. So I proposed a method to help developers who might encounter the issue in future to be able to reference straight from the documentation.

### DIFF
--- a/docs/vrf/v2/best-practices.md
+++ b/docs/vrf/v2/best-practices.md
@@ -104,3 +104,8 @@ If you want to process VRF responses depending on predetermined conditions, you 
 ```solidity
 {% include 'samples/VRF/VRFv2MultiplePaths.sol' %}
 ```
+
+<div class="remix-callout">
+  <a href="https://remix.ethereum.org/#url=https://docs.chain.link/samples/VRF/VRFv2MultiplePaths.sol" target="_blank" >Open in Remix</a>
+  <a href="/docs/conceptual-overview/#what-is-remix">What is Remix?</a>
+</div>
--- a/docs/vrf/v2/best-practices.md
+++ b/docs/vrf/v2/best-practices.md
@@ -97,16 +97,18 @@ function fulfillRandomWords(
 }
 ```
 
-## Processing Different Functions VRF Requests
+## Processing VRF responses through different execution paths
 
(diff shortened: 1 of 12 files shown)


aelmanaa (approved):

franfran20:
@aelmanaa thanks for the review and changes!

docs/vrf/v2/best-practices.md
```diff
@@ -96,3 +96,65 @@ function fulfillRandomWords(
...
+mapping(uint256 => Logic) public requests;
+
+function updateVariableA() public {
+  uint256 requestId = COORDINATOR.requestRandomWords(
+      keyHash,
+      s_subscriptionId,
+      requestConfirmations,
+      callBackGasLimit,
+      1 //numWords
+  );
```
rgottleber, line 120: Would it make sense to pull this into a function instead of repeating it 3 times?

docs/vrf/v2/best-practices.md
```diff
@@ -96,3 +96,65 @@ function fulfillRandomWords(
...
+        override
+  {
+    Logic logic = requests[requestId];
+    if(logic == Logic.A){
+      variableA = randomWords[0];
+    }else if(logic == Logic.B){
+      variableB = randomWords[0];
+    }else if(logic == Logic.C){
+      variableC = randomWords[0];
+    }
```
rgottleber, line 157: This is great, would it be more impactful to call a function instead of just updating a variable when it comes to illustrating that you can _do_ something different based on the call?

docs/vrf/v2/best-practices.md
```diff
@@ -96,3 +96,65 @@ function fulfillRandomWords(
...
+```solidity
+enum Logic { A, B, C}
+
+uint256 public variableA;
+uint256 public variableB;
+uint256 public variableC;
+
+mapping(uint256 => Logic) public requests;
+
+function updateVariableA() public {
```
rgottleber, line 113: Could we create a single `updateVariable` function that takes in the variable to update?

rgottleber (commented):
This is a good example, a few comments to clean things up perhaps.

rgottleber (approved):
LGTM

aelmanaa:
thank you very much @franfran20 for this valuable contribution

franfran20:
You're welcome @aelmanaa ...Happy to contribute anyway I can!

PR #878, opened 2022-07-28
Zoom in images on click

### LINKED ISSUE #877: [Enhancement]: UX for images - Zoom on click
### What would you like us to improve?

Build a feature to be able to zoom in/out when you click on an image

### How important it is?

nice UX improvement. Especially for diagrams

aelmanaa:
## Closing issues

closes #877

## Description

Some diagrams contain lot of info. For UX perspective, would be nice to allow the users to zoom in when they hover over a specific image


_src/docs.css
```diff
@@ -777,6 +777,16 @@ img {
...
 }
 
+div.article > * >  img {
+  cursor: zoom-in;
+  transition: transform ease-in-out 0.5s;
+} 
+
+div.article > * >  img.zoomed {
+  cursor: zoom-out;
+  transform: scale(2.5);
```
dwightjl, line 787:
```suggestion
  transform: scale(1.75);
```
Is there a way to limit the scaled width to the current with of the window? I looked around but didn't find a simple option. If not, 1.75 or 2.0 seems to work better. 2.5 is too large.

dwightjl (changes requested):

dwightjl (approved):

PR #895, opened 2022-08-04
Update performance config docs for Nethermind

vyzaldysanchez:
This adds a warning about using `GasCap` with a client like Nethermind, to avoid running into gas issues.

### DIFF
--- a/docs/Node Operators/evm-performance-configuration.md
+++ b/docs/Node Operators/evm-performance-configuration.md
@@ -203,7 +203,7 @@ RPCTxFeeCap = 0.0
 ```
 
 > 🚧 If you are using `Nethermind` instead of `go-ethereum` you should be aware that the `gascap` flag behaves differently, resulting in errors like [this one](https://github.com/NethermindEth/nethermind/issues/4307).
-> In this case, just set `gascap` to a high number like `5000000000` or just don't set it at all, to avoid issues.
+> In this case, you can set `gascap` to a very large number like `5000000000`.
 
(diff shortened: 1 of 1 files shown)


jmank88 (approved):

vyzaldysanchez:
We would like this to be out as part of the `1.7.0` release. @dwightjl

docs/Node Operators/evm-performance-configuration.md
```diff
@@ -202,6 +202,9 @@ RPCGasCap = 0
 RPCTxFeeCap = 0.0
 ```
 
+> 🚧 If you are using `Nethermind` instead of `go-ethereum` you should be aware that the `gascap` flag behaves differently, resulting in errors like [this one](https://github.com/NethermindEth/nethermind/issues/4307).
+> In this case, just set `gascap` to a high number like `5000000000` or just don't set it at all, to avoid issues.
```
samsondav, line 206:
```suggestion
> 🚧 If you are using `Nethermind` instead of `go-ethereum` you should be aware that the `gascap` flag behaves differently, resulting in errors like [this one](https://github.com/NethermindEth/nethermind/issues/4307).
> In this case, you can set `gascap` to a very large number like `5000000000`.
```

vyzaldysanchez:
Holding off making further changes here since we were told that the `gascap` behavior was fixed (here)[https://github.com/NethermindEth/nethermind/issues/4360] and it will be out in their next release.

PR #988, opened 2022-09-16
Reorganize the data feeds documentation

dwightjl:
- Separate types of feeds
- Reorganize the navigation
- Update data feed processing


docs/data-feeds/historical-data.html
```diff
@@ -2,42 +2,42 @@
 layout: nodes.liquid
 section: ethereum
 date: Last Modified
-title: "Historical Price Data"
-permalink: "docs/historical-price-data/"
-whatsnext: {"API Reference":"/docs/price-feeds-api-reference/", "Contract Addresses":"/docs/reference-contracts/"}
+title: "Historical Data"
+permalink: "docs/data-feeds/historical-data/"
```
aelmanaa, line 6: I would put historical-data under docs/data-feeds/price-feeds/historical-data/

Rational:
- With price feeds, you can ether retrieve the current price or the historical data. Hence, both are path of the same functionality (price feeds)
- historical data uses the api-reference of price feed. Again, that shows that historical data is part of price feeds

_data/navigation.js
```diff
@@ -98,102 +98,102 @@ module.exports = {
...
+                url: '/docs/data-feeds/historical-data/',
+              },
+              {
+                title: 'Price Feeds API Reference',
+                url: '/docs/data-feeds/price-feeds/api-reference/',
+              },
+              {
+                title: 'Contract Addresses',
+                url: '/docs/data-feeds/price-feeds/addresses/',
+                children: [
```
aelmanaa, line 118: still to discuss what's the purpose of nested children since they are not supported by the current nav bar

firebase.json
```diff
@@ -568,6 +568,118 @@
         "source": "/avasummit",
         "destination": "/docs/vrf/v2/introduction/",
         "type": 301
+      },
+
+      /* Data feeds reorg */
+      {
+        "source": "/docs/using-chainlink-reference-contracts/",
```
aelmanaa, line 575: should we also add redirects for  `/docs/using-chainlink-reference-contracts` ?

aelmanaa (changes requested):

firebase.json
dwightjl: Both `/docs/using-chainlink-reference-contracts/` and `/docs/get-the-latest-price/` have redirects. What am I missing?

_data/navigation.js
dwightjl: Removed from the nav and replaced in the Addresses page.

firebase.json
dwightjl: Correct. This is complete.

docs/data-feeds/historical-data.html
dwightjl: I was hoping to generecize this content, but I think I can do that in a followup PR. Moved Historical Data to the correct location.

aelmanaa (approved):

PR #1148, opened 2023-02-15
ETH / USD address updated

cse-amarjeet:
The docs mention a BTC / USD address, but everywhere (youtube tutorials and docs) mention that the address is ETH / USD, so I replaced the address from BTC/USD with ETH/USD address.

## Closing issues

...

## Description

...

## Changes

- High level
- changes that
- you made
- ...


dwightjl (approved):
Any of the feeds should work fine, but it is good to align with the videos to help avoid confusion. Thank you for the PR.

PR #1169, opened 2023-02-26
Update installing frameworks page and image

dwightjl:


thedriftofwords (approved):
Should we start changing header titles to sentence case capitalization a little bit a time with each PR, or should that be a separate mega-PR?

dwightjl:
> Should we start changing header titles to sentence case capitalization a little bit a time with each PR, or should that be a separate mega-PR?

Yes! Updated in this PR. Thank you!

PR #1173, opened 2023-02-28
Functions Beta documentation

dwightjl:


ritika-b (approved):

src/pages/chainlink-functions/tutorials/api-use-secrets.md
```diff
@@ -34,7 +34,7 @@ Apply [here](http://functions.chain.link/) to add your EVM account address to th
 ## Tutorial
 
 :::note[Complete the Before you begin section]
```
thedriftofwords, line 36: State the goal directly in the note "title"

```suggestion
:::note[Set up your environment variables]
```

src/pages/chainlink-functions/tutorials/api-use-secrets.md
```diff
@@ -85,7 +85,7 @@ Reading the output of the example above, you can note that the `BTC/USD` price i
 ### Request
 
 :::note[Reminder]
```
thedriftofwords, line 87: Could we single-source the common environment variable setup steps and then tack on any unique steps per tutorial? That could be more trouble than it's worth if we anticipate the underlying tutorial structure will change soon, but I'm not sure about the timing.

thedriftofwords (approved):

dwightjl:
TODO: Add a banner

src/pages/chainlink-functions/tutorials/api-use-secrets.md
thedriftofwords: Or just make it "Reminder" to be consistent with the other ones.

thedriftofwords: Small improvement that's probably not an unrealistic lift for now: Single-source this reminder note itself.

src/pages/chainlink-functions/tutorials/api-use-secrets.md
thedriftofwords: To clarify: if we single-source the setup steps, then we could probably insert them directly into the tutorial. This seems like a good future enhancement.

dwightjl: This notice is only on a couple of tutorials. Let's look into single-sourcing as a followup.

PR #1174, opened 2023-03-01
Functions updates

dwightjl:


pappas999 (approved):
LGTM

src/pages/chainlink-functions/resources/add-functions-to-projects.md
```diff
@@ -173,10 +173,10 @@ You can deploy consumer contracts using the [Chainlink Functions Starter Kit](ht
    const consumerContract = await ethers.getContractFactory(contractName);
    ```
 
-1. Save the file and run the script to deploy your contract.
+1. Save the file and run the script to deploy your contract. Include the `--network` flag to use a network other than the default in the Hardhat config:
 
    ```shell
-   npx hardhat run ./scripts/deploy.js
+   npx hardhat run ./scripts/deploy.js --network your_network
```
thedriftofwords, line 179: Is this another spot where we should show how to pass Polygon Mumbai as the network specifically? What's the Hardhat default? Do we have a link to hardhat docs that show what these args should look like?

thedriftofwords (approved):

src/pages/chainlink-functions/resources/add-functions-to-projects.md
dwightjl: The default is set in the hardhat config listed here: https://docs.chain.link/chainlink-functions/resources/add-functions-to-projects#configure-hardhat-1

The problem is if someone uses the config above while trying to also use Sepolia. This just reminds devs to override the default if they need to.

PR #1198, opened 2023-03-14
force linkcheck to check the sitemap

aelmanaa:


dwightjl (approved):

dwightjl:
This works locally and the code looks good. I think there are some further improvements we can make to run this more efficiently, but let's make those changes in a followup.

PR #1231, opened 2023-04-04
remove any-api data providers

aelmanaa:


src/features/redirects/redirects.json
```diff
@@ -5,21 +5,6 @@
...
-    {
-      "source": "docs/any-api/data-providers/google-weather",
-      "destination": "any-api/data-providers/google-weather",
-      "statusCode": 301
-    },
-    {
-      "source": "docs/any-api/data-providers/introduction",
-      "destination": "any-api/data-providers/introduction",
-      "statusCode": 301
-    },
```
dwightjl, line 22: Let's update these redirects to go to the Any API intro. Also, add new redirects for the current URLs. We want to avoid 404s.

dwightjl (changes requested):

PR #1252, opened 2023-04-18
CL Nodes 2.0.0 release

dwightjl:


snehaagni (approved):

src/pages/chainlink-nodes/configuring-nodes.mdx
```diff
@@ -71,7 +71,7 @@ You can generate the `config.toml` file using the `chainlink config dump` comman
 
 After you have the `config.toml` file and the `secrets.toml` files on the host system, you can validate these files using a temporary Docker container.
 
-1. Validate the configuration:
+1. Validate the configuration by running the `config validate` command in the Docker container. This command changes to `node validate` when you upgrade your node to version `2.0.0` or later.
```
thedriftofwords, line 74: Could you use the code tabs here to show both commands, showing 2.0.0+ as the default on the left side?

src/pages/chainlink-nodes/v1/running-a-chainlink-node.mdx
```diff
@@ -171,20 +171,20 @@ This guide will teach you how to run a Chainlink node locally using [Docker](#us
      _never_ do this on a production node.
    </Aside>
 
-1. Start the Chainlink Node. Now you can run the Docker image. Replace `<version>` with your desired version. Tag versions are available in the [Chainlink docker hub](https://hub.docker.com/r/smartcontract/chainlink/tags). _The `latest` version does not work._
+1. Start the Chainlink Node. Now you can run the Docker image. Change the version number in `smartcontract/chainlink:2.0.0` with the version of the Docker image that you need to run. For most new nodes, use version `2.0.0` or later. Tag versions are available in the [Chainlink docker hub](https://hub.docker.com/r/smartcontract/chainlink/tags). _The `latest` version does not work._ Chainlink Nodes running `2.0.0` and later require the `-config` and `-secrets` flags after the `node start` part of the command. Versions `1.13.1` and earlier require the `-config` and `-secrets` flags to be placed before `node start`.
```
thedriftofwords, line 174: In this case, nesting tabbed sections would get too unruly, but it would be clearer to show both the 1.x and 2.x versions of each command as code snippets, reducing the slight friction of having people translate this written description to code.

Do we have a way to put the <=1.13.1 version of the commands within a  "click to expand" element below?

thedriftofwords (changes requested):

thedriftofwords (approved):
Approving because that feedback should be non-blocking

src/pages/chainlink-nodes/configuring-nodes.mdx
dwightjl: This section is specific to updating the config to TOML, which can only be done on v1.13.x. The other docs are updated to use the new syntax. The migration doc is the one exception where we need to mention the old commands.

src/pages/chainlink-nodes/v1/running-a-chainlink-node.mdx
dwightjl: I added two extra tabs with v1.13.1 hard coded and the correct syntax for that version. That makes it easier for people running older versions.

PR #1261, opened 2023-04-21
Migrate node ops files to mdx

thedriftofwords:


src/pages/chainlink-nodes/v1/running-a-chainlink-node.mdx
thedriftofwords, line 1: I added this entire file to `.prettierignore`. The regular `prettier-ignore` statement wasn't working on the tabbed code examples starting at line 94, due to line breaks within the code examples.

Alternatively, we could refactor this by putting those tabbed code snippets into another file and then importing it here, but I'd rather do that separately so it doesn't block this PR

thedriftofwords: Other notes: 

- Range ignore is only supported in Markdown, so I removed that.
- I had to put the code tab titles all back as one-liners. Breaking them up didn't fix the issue, and it caused the button CSS issue we've seen before.

dwightjl (approved):

PR #1300, opened 2023-05-15
Chore/doc 150/remove vercel redirects

fmonper1:
Currently we handle redirection in two ways. 
1. The vercel config
2. Meta refresh pages generated during build time

This PR removes the redirects from the vercel config so that there is a single source of truth for our redirects and also introduces middleware for handling redirections at the edge.

Middlewares allow us to keep track of the page referrer by extracting the `referer` header from the http request and appending it to the destination url.

Test url: https://documentation-git-chore-doc-150remove-verc-960820-chainlinklabs.vercel.app/docs/chainlink-keepers/introduction


package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "astro dev",
+    "dev:vercel": "vercel dev",
```
fmonper1, line 7: this allows us to test the middleware locally without deploying to vercel

src/pages/[...redirect].astro
Comment (fmonper1) [inline src/pages/[...redirect].astro:1]: If the docs contains internal links to old urls we should consider keeping this to avoid 404's when developing

vercel.json
```diff
@@ -10,397 +10,5 @@
       ]
     }
   ],
-  "cleanUrls": true,
-  "redirects": [
```
dwightjl, line 14: Are these all duplicates with the Astro redirects? We might need to merge this list into the Astro redirects file and clean up duplicate entries.

fmonper1: Yes, and we shouldn't be using vercel redirections at all with astro 2.0 https://vercel.com/docs/frameworks/astro#can-i-use-vercel.json-to-create-rewrites-with-astro

dwightjl:
@fmonper1 Builds are working again after generating a fresh package-lock.json file but the new linkchecker is throwing errors. I'll take a look and see if I can figure out why.

Spunsd:
Hello

PR #1310, opened 2023-05-18
Add optional copy-paste links for Functions docs

dwightjl:


zeuslawyer (commented):
@dwightjl 
May be worth calling out at the top of this document that we will use curl to fetch files and files may come from two _different repos_: the hardhat starter kit & the functions examples repos (and link to them).  Then we can say something like:
>:warning. _When running your curl commands please pay special attention to the repo that the file is being fetched from as it is possible that both repos may have similar filenames_

Or something like that! Called out early, we arm them with awareness and teach them to pay attention to what they're pasting.

dwightjl:
> @dwightjl May be worth calling out at the top of this document that we will use curl to fetch files and files may come from two _different repos_: the hardhat starter kit & the functions examples repos (and link to them).

I'm hesitant to add a callout, but added some additional clarification and bold text to help distinguish the repos from one another. It's better to have the distinction inline with the instructions if possible so people get the right instructions at the time they are executing them.

thedriftofwords (approved):
Consider placing the new "optionally" statements about copy-pasting _after_ the `curl` commands - in my opinion, it might be easier to parse visually.

dwightjl:
> Consider placing the new "optionally" statements about copy-pasting _after_ the `curl` commands - in my opinion, it might be easier to parse visually.

Good suggestion. This works much better.

PR #1343, opened 2023-06-16
Add translation selector

dwightjl:
The translation selector will not show up until enabled on the API side.


src/components/HeadCommon.astro
```diff
@@ -41,3 +41,34 @@ import "@chainlink/design-system/global-styles.css"
...
+        {
+          button_style: {
+            full_name: false,
+            with_name: true,
+            is_dropdown: true,
+            with_flags: false,
+            flag_type: "circle",
+          },
+          location: {
+            target: "body > header",
```
fmonper1, line 66: Update the header so that there is an empty div next to the github icon
CleanShot 2023-06-16 at 09 18 06@2x

Then you can update the location target so that the dropdown gets inserted inside the navbar

dwightjl: Done

ritika-b (approved):

src/layouts/LandingLayout.astro
```diff
@@ -94,6 +94,7 @@ const canonicalURL = new URL(canonicalURLStr)
       @media (min-width: 50em) {
         .search-item {
           visibility: hidden !important;
+          width: 0 !important;
```
fmonper1, line 97: Why not `display: none`?

fmonper1 (approved):

danielgruesso (approved):

src/layouts/LandingLayout.astro
dwightjl: Done

Issue #1398, opened 2023-07-19
[Enhancement]: Tabs component for switching content 


simkasss:
### What would you like us to improve?

Add a Tabs component for content switching within different sections of a webpage. 
I am looking to create a sample component for content switching.

### How important it is?

A well-designed tabs component would provide intuitive and easy navigation between different sections of a webpage.

PR #1399, opened 2023-07-19
Tabs component for switching content

simkasss:
## Relates issues

relates #1398 

## Description

An example of a Tabs component, designed for content switching within sections of a page.

## Changes

A new TabsContent component was created, based on the existing Tabs component and chainlink/design-system.

Current Tab design:
Screenshot 2023-07-19 at 16 42 06

Suggested Tab design for content switching:
Screenshot 2023-07-19 at 16 41 42


dwightjl (approved):

dwightjl:
Rebased and force-pushed to resolve Git confusion about imports.

dwightjl:
@simkasss I noticed an issue where the tabs do not stay selected in some situations. To reproduce the issue:

1. The first time you load the page, click a "Direct Funding" tab.
2. Scroll down the page until a tab is no longer visible.
3. Scroll back up. The tab selection reverts to Subscription Funding.

It seems like this happens twice, and then the third time you try to cause the issue it no longer happens. Can you reproduce the issue and propose a fix?

simkasss:
@dwightjl, I have identified the issue, I am still working on figuring out an appropriate solution.

src/components/Tabs/useTabState.ts
```diff
@@ -9,12 +9,6 @@ export function useTabState(initialCurr: string, storeKey?: string): [string, (c
   // This ensures a nice SSR result, without waiting for the store to get set on the client!
   const [curr, setCurr] = useState<string>(initialCurr)
 
-  // If you're using a storeKey, set the shared store to an initial value
-  useEffect(() => {
-    if (storeKey) {
-      tabStore.setKey(storeKey, { curr: initialCurr })
-    }
-  }, [storeKey, initialCurr])
```
simkasss, line 17: @dwightjl , the issue is that this code resets the current value to the initial value when the next component renders. If the shared store needs to be set to the initial value, I can add "&& $tabStore[storeKey]?.curr == null" to the condition, and the code works fine (the store does not reset when the user sees the next component).

However, I don't understand why the initial value of the store needs to be set. I deleted this part of the code. Is there any reason why this code was there in the first place?

dwightjl (approved):

src/components/Tabs/useTabState.ts
dwightjl: Looks like this is not necessary. It might have been part of an effort to allow linking to specific tabs with a query parameter, but that capability was never added from what I can tell.

PR #1415, opened 2023-07-25
Fix issue where supported network link did not work for certain netwo…

yacine-benichou:
## Description

In https://docs.chain.link/vrf/v2/estimating-costs, when clicking on the supported networks link after selecting a network, the link does not directly go to the network chosen by the user. Also, There are some typos for the subscription method

## Changes

- Fixed supported networks link issue.
- Fixed typos.


dwightjl (approved):
Approved with one writer edit.

PR #1417, opened 2023-07-26
don't pretty ignore running a chainlink node

aelmanaa:


aelmanaa:
Hello @simkasss 

Check this PR, please. The development experience is a bit terrible. (you need to put ` {/* prettier-ignore */}` on the right spot) but couldn't find a better solution. This is better than ignoring the file

Issue #1423, opened 2023-07-31
[Enhancement]: Extract the dropdown menu from VRF v2 cost calculator


simkasss:
### What would you like us to improve?

The dropdown menu can be extracted from the VRF v2 cost calculator in order to make it reusable in other components.

### How important it is?

Making components reusable improve code consistency and enhance the development experience.

PR #1424, opened 2023-07-31
Extract the dropdown menu from VRF v2 cost calculator

simkasss:
## Relates issues

relates #1423 

## Description

The dropdown menu is extracted from the VRF v2 cost calculator in order to make it reusable in other components.

### DIFF
--- a/src/features/vrf/v2/components/CostTable.tsx
+++ b/src/features/vrf/v2/components/CostTable.tsx
@@ -1,13 +1,13 @@
 import { network, vrfChain } from "~/features/vrf/v2/data"
 import "./costTable.css"
-import { useEffect, useReducer } from "preact/hooks"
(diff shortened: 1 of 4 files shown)


simkasss:
@dwightjl , to make this component reusable on other pages, we should use the same source of chain data (same JSON files) and the same data types in different pages. I plan to modify the VRF v2 cost calculator to use the same [source](https://github.com/smartcontractkit/documentation/blob/e6de98f2be7da36191522d9efff05d91443ebb73/src/features/feeds/data/chains.ts#L34C18-L34C80) as data feeds. I will also move this JSON file higher (e.g. to src/features/data) so it can be used in all the features. Let me know if it doesn't make sense to you.

simkasss:
Closed in favor of https://github.com/simkasss/documentation/pull/1

PR #1473, opened 2023-08-27
Navigation restructure

dwightjl:
- Make left navigation more consistent between each section
- Move Any API
- Split up Getting Started into each product section
- Update the home page
- Update URL paths
- Add redirects


thedriftofwords (commented):
Nit: The product docs links at the top of the page still look off-center on the [home page](https://documentation-git-documentation-reorg-2023-chainlinklabs.vercel.app/) compared to when you click into one of the product doc sets - in that case, the links are properly centered. If this is about to get fixed by a subsequent front-end update, then disregard.

PR #1484, opened 2023-09-08
Remove search shortcut

dwightjl:


thedriftofwords (approved):

thedriftofwords:
Approved as it solves the bug, but would want @fmonper1 to confirm this doesn't have any other unintended side effects

dwightjl:
> Approved as it solves the bug, but would want @fmonper1 to confirm this doesn't have any other unintended side effects

From my testing this only disables the keyboard shortcut. I'm not seeing any downstream effects from this after testing several pages.

PR #1571, opened 2023-10-10
Content preparation overhaul

ryannelsonsmartcontract:
## Closing issues

Closes [FRONT-4351](https://smartcontract-it.atlassian.net/browse/FRONT-4351)

## Description

This PR introduces a new component called `SectionWrapper`. This component is meant to wrap any client-side elements which create headers that are expected to update the table of contents and / or have sticky properties mirroring other article content. The use of this component simplifies the process for writing out client-side content without following our explicit structure to make it work with the TOC and other systems. Just wrap your content in `SectionWrapper`, pass the required props, and it takes care of it. This has been done for our existing components which use this pattern, such as `FeedList` and `AutomationConfigList`.

In addition, a more robust process has been defined for "preparing" header elements. The script to do this is `prepArticleContent.ts > prepareSections()` and it's run in `PageContent.astro` and `Quickstart.astro`.

## Background

Currently we wrap all content following h1 - h4 tags in section tags [using a script](https://github.com/smartcontractkit/documentation/blob/4303d0c327b310754b99b991d5dfc2c19191cc87/src/components/PageContent/PageContent.astro#L183). h1 and h2 tags also cover h3 and h4 tags, acting as "parent" sections. This works with the hierarchy of the table of contents, where h1 (only used for the title in non-quickstarts pages) and h2 are at the top depth, then h3 and h4 are 2nd and 3rd depth.

There are two reasons it's done like this:
1. Sticky headers (h1 and h2) will stay within the bounds of their parent element. Wrapping the headers and their underlying content in section tags gives the sticky headers a parent element to stay within bounds of. Without those tags, the sticky headers would all just stack onto one another as you scroll down
2. Tracking the currently active section in the table of contents requires us to know which content is currently being shown on screen. There are ways to accomplish this by only tracking the headers themselves, but the issue with this system presents itself for nested (2nd and 3rd depth, h3 and h4) content, as well as when scrolling up. The bottom line is that having the section wrappers help make this system a lot more robust and accurate than tracking the header tags

The way it's done creates a situation with `astro-island` components like [`FeedList`](https://github.com/smartcontractkit/documentation/blob/main/src/features/feeds/components/FeedList.tsx), where the headers and the following content aren't present in the DOM at the time the script is being run, since they are hydrated after the DOM loads due to data fetching. There are only a couple of client components we currently have that work like this, so it was easy enough for me to manually wrap the content following headers in section tags manually, but it posed a problem if someone wasn't familiar with this procedure going forward. We needed a better solution.


src/components/SectionWrapper/SectionWrapper.tsx
```diff
@@ -0,0 +1,52 @@
...
+  idOverride?: string
+  updateTOC?: boolean
+  key?: Key
+}
+
+/**
+ * Use this component to wrap client-side components which feature section headers between h2 - h4.
+ * Provides sticky functionality, anchor tags and updates table of contents as needed
+ */
+function SectionWrapper({ title, depth, children, idOverride, updateTOC = true, key }: Props) {
```
Alejandro-Larumbe, line 21: Suggestion: consider having this false by default. Currently when this wrapper is implemented, only one time this is false, needing `updateTOC=false` the rest of the times.

Alejandro-Larumbe (approved):

src/components/SectionWrapper/SectionWrapper.tsx
ryannelsonsmartcontract: This is a valid point as far as the current implementations go. However, it should be considered default behavior to update the table of contents when a new header is added after the first render, such as in the case of `FeedList.tsx` line 171. By changing the default to false, you risk future implementations where the TOC is not updated when it should be.

aelmanaa (approved):

PR #1578, opened 2023-10-12
Add warning text and icon when gas price is higher than selected gas lane to VRF calculator

yacine-benichou:
## Description

During a gas spike for a specific network, estimated cost in the VRF calculator is no longer valid because the gas price is higher than the current gas lane.

## Changes
Added a warning text when this case happens.

### DIFF
--- a/src/features/vrf/v2/components/CostTable.tsx
+++ b/src/features/vrf/v2/components/CostTable.tsx
@@ -8,7 +8,7 @@ import button from "@chainlink/design-system/button.module.css"
 interface Props {
   method: "vrfSubscription" | "vrfDirectFunding"
   network: string
-  icon: HTMLElement | undefined
+  aside: HTMLElement | undefined
 }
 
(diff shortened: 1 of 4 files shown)


src/features/vrf/v2/components/CostTable.tsx
```diff
@@ -697,6 +703,16 @@ export const CostTable = ({ method, network }: Props) => {
             Calculate
           </button>
         </div>
+        {BigNumber.from(state.currentGasPrice).gt(BigNumber.from(state.currentGasLane)) && (
+          <>
+            <p>
+              <Icon />
```
thedriftofwords, line 709: `Icon` expects a `type` input; try specifying `type="warning"`?

thedriftofwords: Alternatively, use the `Aside` component instead

thedriftofwords (approved):
Looks great!

PR #1585, opened 2023-10-15
Overview sections: consistent org

aelmanaa:


src/config/sidebar.ts
```diff
@@ -314,6 +314,10 @@ export const SIDEBAR: Partial<Record<Sections, SectionEntry[]>> = {
           title: "Supported Networks",
           url: "chainlink-automation/supported-networks",
         },
+        {
+          title: "Billing",
+          url: "chainlink-automation/automation-economics",
+        },
```
ritika-b, line 320: Will have to remove this as it is already updated

ritika-b (approved):

PR #1638, opened 2023-11-14
Chainlink Node Config Fixes

george-dorin:
Add documentation on `NodePool.LeaseDuration` that shipped with version `2.7.0`


dwightjl:
These changes get pulled in from https://github.com/smartcontractkit/chainlink/blob/develop/core/config/docs/chains-evm.toml#L323 when a new version releases. But there's still a small fix we can make in this PR so we'll merge it.

dwightjl (approved):

PR #1651, opened 2023-11-16
Fix language selector

Alejandro-Larumbe:

## Closing issues

...

## Description

...

## Changes

- High level
- changes that
- you made
- ...

### DIFF
--- a/src/components/Header/NavBar.tsx
+++ b/src/components/Header/NavBar.tsx
@@ -5,6 +5,8 @@ import { getNavigationProps } from "./getNavigationProps"
 import { useNavBar } from "./useNavBar/useNavBar"
 import styles from "./scroll.module.css"
 
+declare const Weglot: any
+
 export const NavBar = ({ path, showSearch = true }: { path: string; showSearch?: boolean }) => {
   const navRef = React.useRef(null)
 
--- a/src/components/Header/NavBar.tsx
+++ b/src/components/Header/NavBar.tsx
@@ -24,6 +24,42 @@ export const NavBar = ({ path, showSearch = true }: { path: string; showSearch?:
     }
   }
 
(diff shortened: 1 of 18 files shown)


src/components/Header/NavBar.tsx
```diff
@@ -5,6 +5,8 @@ import { getNavigationProps } from "./getNavigationProps"
 import { useNavBar } from "./useNavBar/useNavBar"
 import styles from "./scroll.module.css"
 
+declare const Weglot: any
```
Alejandro-Larumbe, line 8: Without this TS check fails because doesn't know of the global Weblog variable.

src/components/Header/NavBar.tsx
```diff
@@ -24,6 +26,42 @@ export const NavBar = ({ path, showSearch = true }: { path: string; showSearch?:
     }
   }
 
+  React.useEffect(() => {
+    if (
+      !window.location.hostname.includes("localhost") &&
+      !window.location.hostname.includes("documentation-private-git-")
```
zeosamaster, line 32: Does this mean it's only meant to have have the language bar in prod?
Is there an env var available that allows us not to leak our preview deployment domain?

src/components/Header/NavBar.tsx
danielgruesso: LGTM

src/components/Header/NavBar.tsx
danielgruesso: > Does this mean it's only meant to have have the language bar in prod?
Correct, it's domain dependent.

> Is there an env var available that allows us not to leak our preview deployment domain?
No. We can protect them or remove them, however, since docs are oss it should not be an issue.

dwightjl (approved):

PR #1665, opened 2023-11-22
Add Updated Functions code

ritika-b:
- adds try/catch to Function Automation code example


KuphJr (approved):
Overall, this looks pretty good! However, I wonder if it is worth adding a note somewhere more prominent about the risk of Automation infinitely attempting to perform an upkeep if the last upkeep time is never updated due to a reversion.
Maybe we could add a note or warning closer to the top of [this guide](https://docs.chain.link/chainlink-functions/tutorials/automate-functions-custom-logic) so devs don't miss it?

ritika-b:
> Overall, this looks pretty good! However, I wonder if it is worth adding a note somewhere more prominent about the risk of Automation infinitely attempting to perform an upkeep if the last upkeep time is never updated due to a reversion. Maybe we could add a note or warning closer to the top of [this guide](https://docs.chain.link/chainlink-functions/tutorials/automate-functions-custom-logic) so devs don't miss it?

Adding this! Thanks

PR #1878, opened 2024-04-13
Fix typos

omahs:
Fix typos


aelmanaa (approved):
Thank you very much @omahs !

PR #1881, opened 2024-04-15
allow table to overflow horizontally

aelmanaa:


thedriftofwords:
Before:
Screenshot_20240415-131332

After:
Screenshot_20240415-131356

thedriftofwords (approved):
Massive UX improvement, ty!

PR #2123, opened 2024-10-14
Added native token top-up and withdraw functionality in VRFv2PlusSubscriptionManager contract.

SyedAsadKazmi:
Some devs are asking for this functionality, as can be seen from this [Eth-StackExchange post](https://ethereum.stackexchange.com/questions/165730/chainlink-vrf-2-5-fund-subscription-with-native-token-programmatically/165812#165812), so I've added the`topUpSubscriptionWithNativeToken` and `withdrawNativeToken` functions to support the same.


SyedAsadKazmi:
> Hi @SyedAsadKazmi , thanks for the contribution. Since this PR is to help users to fund the vrf 2.5 subscription with native tokens, I think it might be a good idea that you can also add a new function to allow them to request randomness with native payment.
> 
> For now, although users can fund the subscription with native tokens, there is no function in the contract for them to request randomness with native balance.

Hello @QingyangKong,

Thanks for the feedback. I've just added the `enableNativePayment` argument to the `requestRandomWords()` function. Please check.

QingyangKong (approved):

PR #2205, opened 2024-12-05
Ccip/1.5.1 remix

aelmanaa:


thedriftofwords (commented):
Non-blocking feedback:

Prerequisites
- Web browser setup: Add link or reminder to check supported networks page
- Native gas tokens: Add link to billing page (for users who need guidance on how much gas tokens are enough

Blockchain setup
- UI nit: The "<>" between the dropdown fields is clickable which might lead users to expect they can swap the values of the fields by clicking that

thedriftofwords (approved):
Excellent 🚀

PR #2394, opened 2025-02-20
fix: fix error handling to properly throw error messages

onasilos:
## Description
I noticed an issue with error handling in the code. The line `throw Error(err)` could cause issues if `err` isn't an instance of the `Error` object. I updated it to `throw new Error(err instanceof Error ? err.message : String(err))`, ensuring that error messages are always correctly thrown, even if `err` is a string or other non-`Error` object.

## Changes
Replaced `throw Error(err)` with `throw new Error(err instanceof Error ? err.message : String(err))` to ensure proper error message throwing.

**This should make the error handling more reliable.**


khadni (approved):

khadni:
Thank you for your first contribution and this improvement, @onasilos!

PR #2453, opened 2025-03-15
Update testnet LINK address

sonyschan:
## Closing issues

N/A

...

## Description

Wrong LINK CA

...

## Changes

Update LINK CA to be 0xb1D4538B4571d411F07960EF2838Ce337FE1E80E


gfletcher-cll:
Thank you for your edit! In this case, the current address provided is the correct one for Ethereum Sepolia.  The address you provided is for Arbitrum Sepolia.  Given, we don't specifically call out Ethereum Sepolia in the guide, I will update the guide so no one else gets confused.

PR #2493, opened 2025-03-31
new - Update automation-interfaces.mdx

dedyshkaPexto:
fix old to a new one


gfletcher-cll (changes requested):
thanks for catching this outdated link!  the one you provided however, is just the redirect to the new landing page.  Can you please update with the correct link? https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_call

gfletcher-cll:
That may be the link that is redirected to from the original wiki link, but it is in fact just a fallback URL when they can't find the correct content.  The URL you provided is just the landing page for ethereum vs the actual content for the `eth-call` endpoint.  Please update with the recommended link.  If you prefer not to, I will go ahead and close this PR and update the link separately.  Appreciate you letting us know the link it outdated!

PR #2542, opened 2025-04-11
update the svr searcher documentation and add code snippets in ts/go

KMontag42:
## Closing issues

closes #DOC-994

## Description

Updates the SVR Searcher documentation with several code snippets, extra considerations, and more detailed steps to help onboard searchers more rapidly.

## Changes

- adjusted the numbered sections after `1. High Level Steps`
- added code snippets for several of the new steps
- re-organized existing steps to follow the desired flow more accurately
- added a line item under `considerations`
- added import for `Tags` component so that we can do the tabbed code language selection on the documentation code samples

## NOTE

It would be great if we could make the `samples/DataFeeds/SVR/decoding-abi.json` file a single line file, but the prettier git hook wouldn't let me.

Not the biggest deal if we cannot do that, but it would save some vertical space :)


src/content/data-feeds/svr-feeds/index.mdx
```diff
@@ -183,13 +185,15 @@ Chainlink SVR uses forwarder contracts to route feed updates. You typically see
...
+Once you've identified a potential SVR feed update transaction, you'll need to decode its payload to extract the new price data. Extracting this price information is essential for determining if profitable liquidation opportunities exist. The decoding process involves understanding two function calls:
+
+1. First, the transaction calls the `forward` function:
 
 ```solidity
 forward(address to, bytes callData)
 ```
 
-You can decode the new price from the `forward` `callData` that calls:
+1. Inside the `callData` parameter is the encoded call to `transmitSecondary`, which contains the actual price data:
```
KMontag42, line 196: i think you need to use `2.` here since you are separated from the previous list entry (so the `1. 1. 1.` technique won't work)

khadni: fixed by indenting

KMontag42: image
from the preview link

gfletcher-cll (approved):

PR #2565, opened 2025-04-20
 Incorrect CSS variable name

detrina:
Replace the custom property `--font-family-mono` with the correct `--font-mono` to match the declared variable.


khadni:
Thanks @detrina!

PR #2717, opened 2025-06-08
should be two instead of three

mcmoodoo:

## Closing issues

...

## Description
Simple typo most likely

...

## Changes

- High level
- changes that
- you made
- ...


khadni:
Hey @mcmoodoo, thanks for this suggestion. While only two prices (bid and ask) calculate the spread width, all three prices are needed to "form" the complete pricing spread info that gives protocols insight into market activity. The sentence isn't about mathematical calculation but about the informational structure provided to protocols.

Does it make sense?

PR #3133, opened 2025-11-04
MVR Bid Clarification

dev-dist:
- Added clarification on how to submit bids in Search Guide for Chainlink SVR
- Updated code with comments explaining bid submission

### DIFF
--- a/public/changelog.json
+++ b/public/changelog.json
@@ -1798,6 +1798,13 @@
       "topic": "Data Feeds"
     },
     {
+      "category": "release",
+      "date": "2025-10-01",
+      "description": "[DataLink](https://docs.chain.link/datalink) is now live. View the [Provider Catalog](https://docs.chain.link/datalink/provider-catalog) to explore institutional data available for onchain applications.",
+      "title": "Chainlink DataLink",
+      "topic": "General"
+    },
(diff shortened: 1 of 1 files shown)


public/samples/DataFeeds/SVR/broadcaster.go
```diff
@@ -1,6 +1,8 @@
 import "github.com/flashbots/mev-share-node/mevshare"
 
 // ... create and sign your transaction
+// Your transaction should send ETH to block.coinbase as the bid
```
KMontag42: i don't think we should put this info into the go code itself as it has to be done in solidity.

public/samples/DataFeeds/SVR/broadcaster.ts
```diff
@@ -14,6 +14,8 @@ async function sendBackrunBundle(pendingTx: IPendingTransaction) {
 }
 
 async function buildDummyBackrunTx(pendingTx: IPendingTransaction) {
+  // Your liquidation transaction should send ETH to block.coinbase as the bid
```
KMontag42: same here, i don't think we should put this info into the typescript since it needs to be done in solidity.

PR #3191, opened 2025-11-19
fix: update hyperlink URL path to resolve 404 error

bsilagani:
Fixed the hyperlink URL path that was causing a 404 error. Updated it to redirect to the correct page.


khadni:
https://github.com/smartcontractkit/documentation/pull/3193

PR #3879, opened 2026-06-15
Update TON native currency from TON to GRAM

SyedAsadKazmi:


Genesis3800 (approved):
Approving at 1200 hours UTC

PR #4054, opened 2026-08-12
Add missing Flags contract addresses to registry

sebastian-lerner-link:
## Summary
- Adds 7 chains to the Flags contract registry table that have deployed Flags contracts but were missing from this page: Hyperliquid, MegaETH, Plasma, Polygon Katana, Hedera, Metis, and X Layer.
- Addresses sourced from `reference-data-directory` (`flags/` dirs per chain), cross-checked against the existing registry format/pattern.

## Test plan
- [ ] Verify each address resolves correctly on its linked explorer
- [ ] Confirm the table renders correctly on the docs site preview


robin-rrt (commented):
some of these contracts are not verified on block explorer, I would get them verified as soon as we can @sebastian-lerner-link

PR #4173, opened 2026-09-14
Add TT and PTT Canton tutorials back with a note on LINK token finality

SyedAsadKazmi:


src/content/ccip/tutorials/canton/destination/token-transfers.mdx
```diff
@@ -55,6 +55,7 @@ Key points:
 - **Burn and mint**: TEST is burned on Sepolia; LINK is minted on Canton after execution.
 - **Manual execution required**: Unlike `canton2any` default auto-exec on Sepolia, all `any2canton` token transfers require `any2canton:manual-exec` on Canton.
 - **Token pool disclosures**: Execution fetches disclosures from Global CCIP EDS and the token pool operator EDS.
+- **Finality**: The TEST ↔ LINK token pool only supports default (`finalized`) source finality. Unlike data-only sends, you cannot pass `--finality` to `any2canton:token`.
```
Giuliano-1: This isn't so clear to me. As in its not true you can't pass `--finality` to `any2canton:token`. Its just that you don't do it for LINK. Also seems that this should be checked by the SDK, like if finality must be reached that can be validated from the TP and then the flag is just rejected. Something to raise to Andre.

SyedAsadKazmi: Yeah, makes sense. This check needs to be implemented at the SDK-level, to ensure whether the dest TP supports custom finality, before proceeding from the source, as otherwise the tx still gets submitted on source but stuck on manual exec.

Giuliano-1 (approved):

PR #4179, opened 2026-09-16
Add `arc-mainnet` and directory refresh

SyedAsadKazmi:


Genesis3800 (approved):
Regular directory update, adding Arc-Mainnet

